PR #210, opened 2024-04-23
feat: use rayon::scope for threaded_receive

QuarticCat:
A workaround for a workaround. How cursed!

As rayon already exists in typst, there's no extra dependency introduced.

The time of `./target/release/tinymist lsp --replay target/e2e/tinymist/vscode/mirror.log` reduced from ~3.7s to ~3.0s on my machine, measured by `hyperfine --warmup 3`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3803,6 +3803,7 @@ dependencies = [
  "once_cell",
  "parking_lot",
  "paste",
+ "rayon",
  "serde",
  "serde_json",
  "tinymist-query",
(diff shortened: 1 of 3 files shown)


Myriad-Dreamin:
If the performance is fixed by #211, we can close it. We would be better to improve the LSP itself than improve the workaround function. Furthermore, this PR adds extra constraint to the `threaded_receive`, that it may consume all rayon threads when some `threaded_receive`s depend on other `threaded_receive`s.

QuarticCat:
As #211 has been merged, I think it's fair to close this PR.

PR #222, opened 2024-05-03
feat: conditionally open activity icon when lang id is typst

Enter-tainer:
Hide activity bar icon when res lang id is not typst

see 
- https://github.com/Enter-tainer/typst-preview/pull/288
- https://github.com/microsoft/vscode/issues/49145


Myriad-Dreamin:
Thank you!

PR #271, opened 2024-05-11
dev: interning types

Myriad-Dreamin:
This slows down a bit, but we will rescue the performance after finishing type system.


crates/tinymist-query/src/adt/interner.rs
```diff
@@ -0,0 +1,283 @@
...
+                    .clone(),
+            },
+        }
+    }
+}
+
+// Note: It is dangerous to keep interned object temporarily (u128)
+// Case:
+// ```
+// insert(hash(Interned::new_str("a"))) == true
```
Myriad-Dreamin, line 68: Should be
```
// insert(hash(Interned::new_str("a"))) == true
// insert(hash(Interned::new_str("a"))) == true
```

PR #314, opened 2024-05-23
Update Cargo.toml

sbromberger:
Strips binary on release, moves debug to dev. This saves half a gigabyte on release builds by automatically stripping the binary.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -99,6 +99,9 @@ tinymist = { path = "./crates/tinymist/" }
 tinymist-query = { path = "./crates/tinymist-query/" }
 tinymist-render = { path = "./crates/tinymist-render/" }
 
+[profile.dev]
+debug = true
+
 [profile.dev.package.insta]
 opt-level = 3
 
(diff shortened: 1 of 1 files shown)


Myriad-Dreamin:
The debug info should be spilt by external tools and will be released and used for debugging tools like gd .
https://github.com/Myriad-Dreamin/tinymist/blob/9d4700aee3ec09eaf0515c3dc87436897aa03c92/.github/workflows/release.yml#L124

sbromberger:
Debug info shouldn't really be in a `--release` build though. This way, if you just do a `cargo build`, you'll get the symbol info (and a half-gig binary), but if you do a `--release`, you will get something optimized for production (48 MB).

Enter-tainer:
> @Enter-tainer idea on managing profiles?

https://github.com/Enter-tainer/typstyle/blob/master/Cargo.toml#L96-L96

the idea is, keep rust's default profiles untouched(debug, release..) and add a new one that inherit them.

PR #319, opened 2024-05-30
feat: generate names for destructuring closure params

wrenger:
This PR generates parameter names for destructured patterns, as shown below.

```typ
#let foo((x, y, _, ..z)) = [ #x #y #z ]
#foo((1, 2, 3, 4, 5, 6))
```


Myriad-Dreamin:
Seems like we just work around the problem by stringifying the entire pattern, but it is totally fine currently since we haven't done anything with binding names in patterns. I would like to test whether it works well with `textDocument/signatureHelp` before merging.

Myriad-Dreamin:
🚀

PR #320, opened 2024-05-31
refactor: simplify pin_entry

QuarticCat:
The async version cannot steal twice.


crates/tinymist/src/state.rs
```diff
@@ -30,17 +30,13 @@ impl CompileServer {
 impl TypstLanguageServer {
     /// Pin the entry to the given path
     pub fn pin_entry(&mut self, new_entry: Option<ImmutPath>) -> Result<(), Error> {
-        self.pinning = new_entry.is_some();
```
Myriad-Dreamin, line 33: I see `self.pinning` is not updated now, so we need to update `focus_entry` and other related functions as well in this PR.

QuarticCat:
All resolved.

PR #349, opened 2024-06-27
fix: avoid duplicated method completion

Myriad-Dreamin:


laurmaedje:
doesn't this drop purely mutable methods like array.push? (compared to methods which can be both like array.at)

Myriad-Dreamin:
@laurmaedje I'm not confident on whether it is absolutely correct to just remove the code for fix. But we are still completing `array.push`.

image

PR #365, opened 2024-07-03
feat: support querying label with paper name in bib items

kririae:
Record the progress here. Suppose we do have three labels, with VSCode as the LSP client 
1. `<jiang2015apic>`
2. `<jiang2016mpm>`
3. `<bridson2008book>`

Type, for example, `#cite(<book`, [3] will correctly pop out because it is prefixed with `b`. However when we're searching for `apic`, `#cite(<apic` will give a empty completion list, because the corresponding label is not prefixed with `a`.

Although we can mitigate this through setting `editor.suggest.matchOnWordStartOnly` in VSCode to false, it is not the expected default behavior. Current solution is to make use of the [filterText](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_completion) option, by concatenating the "{label}|{bib_title}" sections, the behavior changes to: 
1. typing `#cite(<apic` will correct give the match to [1]
2. searching for any word inside the paper title will also give the corresponding completion item.

Note that completion sorting and filtering is still provided by the client.


yarn.lock
```diff
@@ -3659,7 +3675,7 @@ typescript@^5.2.2, typescript@^5.3.3:
     "@myriaddreamin/typst-ts-renderer" "0.5.0-rc5"
     "@myriaddreamin/typst.ts" "0.5.0-rc5"
     rxjs "^7.8.1"
-    typst-dom "link:../../../AppData/Local/Yarn/Cache/v6/npm-typst-preview-frontend-0.0.0-e4b78602-ab80-42c0-9a76-2c66cb729491-1719528135321/node_modules/typst-dom"
```
Myriad-Dreamin, line 3662: we can discard changes of yarn.lock.

crates/tinymist-query/src/upstream/complete.rs
```diff
@@ -1119,9 +1130,10 @@ impl<'a, 'w> CompletionContext<'a, 'w> {
                         if close { ">" } else { "" }
                     )
                 }),
-                label_detail: label_desc,
                 label: label.as_str().into(),
+                label_detail: label_desc,
                 detail,
+                filter_text: Some(filter_text),
```
Myriad-Dreamin, line 1136: I have tried another strategy, that separates another completion item for each bib title, and think of it would be better, since vscode will hightlight pattern matched in title for you:

image

image

```rs
let label: EcoString = label.as_str().into();
let completion = Completion {
    kind: CompletionKind::Constant,
    apply: (open || close).then(|| {
        eco_format!(
            "{}{}{}",
            if open { "<" } else { "" },
            label.as_str(),
            if close { ">" } else { "" }
        )
    }),
    label: label.clone(),
    label_detail: label_desc,
    filter_text: Some(label.clone()),
    detail,
    ..Completion::default()
};

if let Some(bib_title) = bib_title {
    self.completions.push(Completion {
        label: bib_title.clone(),
        label_detail: Some(label),
        filter_text: Some(bib_title),
        ..completion.clone()
    });
};
self.completions.push(completion);
```

crates/tinymist-query/src/upstream/complete.rs
```diff
@@ -1107,8 +1109,17 @@ impl<'a, 'w> CompletionContext<'a, 'w> {
             label,
             label_desc,
             detail,
+            bib_title,
         } in labels.into_iter().skip(skip).take(take)
         {
+            let filter_text = if let Some(bib_title) = bib_title {
+                // now format it like this, this accidentally fixes the issue
+                // with the `Match On Word Start Only` option in vscode.
+                eco_format!("{}|{}", label.as_str(), bib_title)
```
Myriad-Dreamin, line 1118: not sure whether `filter_text` really fixes the problem, since `filter_text` is set to `label` by default.

yarn.lock
kririae: I'll remove this change in the next commit

PR #374, opened 2024-07-07
build: add vscode launch script for typst-preview

Enter-tainer:


yarn.lock
Enter-tainer, line 1: After I run `yarn` the lockfile changes for some reason. IDK if this is excepted

Myriad-Dreamin: looks no problem

PR #445, opened 2024-07-21
Fix: Handling unwrap for the args in compile command

upsidedownsweetfood:
I was simply bothered by how the program just panicked if I did not include a path in the compile command, so i replaced the unwrap with a match statement.

**before:**
image

**after:**
image


Myriad-Dreamin:
This is an internal command though. I would like to hide it in a later PR.

PR #454, opened 2024-07-24
feat(typlite): add equation support

Myriad-Dreamin:


crates/typlite/src/lib.rs
```diff
@@ -252,6 +250,33 @@ impl TypliteWorker {
...
+        let mut tracer = Tracer::default();
+        let document = typst::compile(&world, &mut tracer)
+            .map_err(|e| format!("compiling math node: {e:?}"))?;
+
+        let svg_payload = typst_svg::svg_merged(&document, Abs::zero());
+        let base64 = base64::engine::general_purpose::STANDARD.encode(svg_payload);
+
+        if inline {
+            Ok(Value::Content(eco_format!(
+                r#"<img style="vertical-align: -0.35em" src="data:image/svg+xml;base64,{base64}" alt="typst-block" />"#
```
Enter-tainer, line 271: ideally, we should obtain baseline position like what rehype-typst did

Myriad-Dreamin: I know it, but I've do too much work in this PR, so a harden value of alignment should be acceptible. This magic alignment should only fit in "New Computer Modern Math", and we currently haven't supported fonts customized yet, so everything should just work.

Enter-tainer: > the trick was actually copied from rehype-typst.

I suppose you mean that the `vertical-align` css trick is copied from rehype-typst. But the offset itself is not
https://github.com/Myriad-Dreamin/typst.ts/blob/main/projects/rehype-typst/lib/index.js#L118

Enter-tainer: > This magic alignment should only fit in "New Computer Modern Math", and we currently haven't supported fonts customized yet, so everything should just work.

...as long as there is no large inline equations like fraction towers

PR #458, opened 2024-07-25
feat: local package manager

OrangeX4:
1. Auto complete for `#import "@local/xxx"`
2. Create or open a local package in tinymist


crates/tinymist-query/src/upstream/complete.rs
```diff
@@ -1042,9 +1043,73 @@ impl<'a, 'w> CompletionContext<'a, 'w> {
         }
     }
 
+    /// Get local packages
+    fn local_packages(&mut self) -> Vec<(PackageSpec, Option<EcoString>)> {
+        // search packages locally. We only search in the data
+        // directory and not the cache directory, because the latter is not
+        // intended for storage of local packages.
+        let mut packages = vec![];
+        if let Some(data_dir) = dirs::data_dir() {
```
Myriad-Dreamin, line 1052: could you use some trick to unwrap this huge nested block?

For example:
```rs
let Some(data_dir) = dirs::data_dir() else {
  return packages;
}

if !local_path.exists() {
  return packages;
}
```

editors/vscode/src/package-manager.ts
```diff
@@ -0,0 +1,209 @@
+import { window, workspace } from 'vscode';
+import * as fs from 'fs';
+
+// error message
+export const dataDirErrorMessage = 'Can not find package directory.';
+
+// https://github.com/typst/packages#local-packages
+export function getDataDir() {
```
Myriad-Dreamin, line 8: The `process.env` uses looks not great to me. I would like to use same configuration as that from LSP, which already computes the directory to local packages.

Could you add a resource that provides data directory at the LSP side? 

Specifically, we can add a resource `dirs/local-package` here, which returns same value as `function getLocalPackagesDir` in this file:

https://github.com/Myriad-Dreamin/tinymist/blob/96d5ba0c0075072b4e602dd6291b8f1d1c2de0be/crates/tinymist/src/server.rs#L266-L269

Then, we can get the resources by `tinymist.getResource`:

https://github.com/Myriad-Dreamin/tinymist/blob/96d5ba0c0075072b4e602dd6291b8f1d1c2de0be/editors/vscode/src/lsp.ts#L18-L21

https://github.com/Myriad-Dreamin/tinymist/blob/96d5ba0c0075072b4e602dd6291b8f1d1c2de0be/editors/vscode/src/lsp.ts#L32-L34

Myriad-Dreamin (commented):
LGTM

OrangeX4:
ready.

Myriad-Dreamin:
🎉

PR #464, opened 2024-07-26
feat(invert-colors): add normal-image option

SetsuikiHyoryu:


Myriad-Dreamin:
https://github.com/Enter-tainer/typst-preview/pull/301

tools/typst-preview-frontend/src/ws.ts
```diff
@@ -357,7 +357,8 @@ export async function wsMain({ url, previewMode, isContentPreview }: WsArgs) {
                 svgDoc.setPartialRendering(true);
                 return;
             } else if (message[0] === "invert-colors") {
-                const strategy = dec.decode((message[1] as any).buffer);
+                const rawStrategy = dec.decode((message[1] as any).buffer).trim();
+                const strategy = INVERT_COLORS_STRATEGY.find(t => t === rawStrategy) || (JSON.parse(rawStrategy) as StrategyMap);
```
Myriad-Dreamin, line 361: I fixed a bug about type (deserialization). Previously, the input is just raw string without deserialization.

tools/typst-preview-frontend/src/ws.ts
```diff
@@ -393,30 +394,40 @@ export async function wsMain({ url, previewMode, isContentPreview }: WsArgs) {
...
+    const decide = (strategy: StrategyKey) => {
+        switch (strategy) {
+            case 'never': default: return false;
+            case 'auto': return (autoDecision ||= { value: determineInvertColor() }).value;
+            case 'always': return true;
+        }
     }
 
+    root.classList.toggle("invert-colors", decide(strategy?.rest || 'never'));
+    root.classList.toggle("normal-image", !decide(strategy?.image || strategy?.rest || 'never'));
```
Myriad-Dreamin, line 429: There are some subtle changes here.
- if the strategy of `image` is not specified, it inherits value from "rest" selector.

Myriad-Dreamin (commented):
I have fixed some bugs and refined the idea a bit. I did:
- only allow input either by a string or a strategy (object) map.
- change "selector" to elements from all to rest.
- if the strategy of image is not specified, it inherits value from "rest" selector.

Myriad-Dreamin:
I ran a bit and didn't find any problem. Could you run the extension locally and confirm this PR solving your problem exactly?

SetsuikiHyoryu:
@Myriad-Dreamin 
I successfully started the debug function using F5, but I'm not sure how to configure parameters in this state. Configuring parameters as I would normally set up a plugin doesn't seem to work.
image

SetsuikiHyoryu:
@Myriad-Dreamin 
I successfully started the debugging feature using F5, but it seems like the configuration parameters didn’t take effect according to the new code. Is it possible that I ran `yarn build` in the wrong location?
image
image

Myriad-Dreamin:
F5 starts a LSP extension, which changed the name of `typst-preview.xxx` to `tinymist.preview.xxx`, such as `tinymist.preview.invertColors`

SetsuikiHyoryu:
@Myriad-Dreamin 
I deleted my local directory and recloned it, which fixed the issue. I suspect the problem might have been related to caching in my node_modules directory.
image

PR #465, opened 2024-07-26
fix: Modifies static host to send Content-Type: text/html

### LINKED ISSUE #463: No Content-Type header on static html
I was testing using this in github Codespaces and the browser just showed the text of the HTML.  After some debugging, I noticed that `Content-Type` header is not included in the response.  The static web server should probably send responses with `Content-Type` set to `text/html`.

After browsing through the code a bit, it looks like `hyper::Response::new` [is used](https://github.com/Myriad-Dreamin/tinymist/blob/22aea0e0752494c52c14d1503004a311b549b144/crates/tinymist/src/tool/preview.rs#L372C64-L372C67).

...instead of `Response::builder()`, which might look something like

```
let response = Response::builder()
    .header("Content-Type", "text/html")
    .header("Content-Length", body.len())
    .body(body.into())
    .unwrap();
```

cskeeters:
Resolves #463 

Disclaimer: these are my first lines of Rust ever.  It appears to work.

Thanks for having a straightforward `CONTRIBUTING.md` file.  I ran `./scripts/e2e.sh` successfully.  At first, `cargo insta` failed, but I figured out what it was and installed it.  A link to [cargo-insta](https://crates.io/crates/cargo-insta) might help other first-timers.

### DIFF
--- a/crates/tinymist/src/tool/preview.rs
+++ b/crates/tinymist/src/tool/preview.rs
@@ -369,15 +369,18 @@ pub fn make_static_host(
                 async move {
                     if req.uri().path() == "/" {
                         log::info!("Serve frontend: {:?}", mode);
-                        Ok::<_, hyper::Error>(hyper::Response::new(hyper::Body::from(html)))
+                        let res = hyper::Response::builder()
+                            .header(hyper::header::CONTENT_TYPE, "text/html")
+                            .body(html)
+                            .unwrap();
+                        Ok::<_, hyper::Error>(res)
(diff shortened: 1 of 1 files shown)


Myriad-Dreamin:
Looks pretty well.

> A link to [cargo-insta](https://crates.io/crates/cargo-insta) might help other first-timers.

An error complaining missing command will guide you to search and find a way to install it, but it should be better to provide a link to help install. lets improve it right now.

Issue #483, opened 2024-08-01
Label completion using `@` replaces with reference title instead of label


quachpas:
Hello! Thanks for all the work

**Describe the bug**
When auto-completion triggers with character `@` for a reference, the replacement text is the title instead of the label.
![screenshot-2024-08-01-18-19-17](https://github.com/user-attachments/assets/5d7573ae-51cc-41b8-b03d-14431f055af0)

With `#cite`, there are no issues and the query replacement is the bibliography item label.

Is this intended behaviour? In which case, would it be possible to limit the reference title completion suggestion to `#cite` and not `@`?

Thanks!
 
**Package/Software version:**

VSCode version(Help -> About):
```plain
Version: 1.92.0-insider
Commit: 4d17e85cfd911fb6b116e08a4d7c37cd725a966a
Date: 2024-07-26T13:12:28.318Z
Electron: 30.1.2
ElectronBuildId: 9870757
Chromium: 124.0.6367.243
Node.js: 20.14.0
V8: 12.4.254.20-electron.0
OS: Linux x64 6.10.2-arch1-1
```

tinymist extension version: `v0.11.0`. Get it by `tinymist --version` in terminal.
```plain
tinymist 
Build Timestamp:     2024-07-27T10:27:48.450267030Z
Build Git Describe:  v0.11.17-dirty
Commit SHA:          651d7f7148448e815d8e50c03e52d48f714f85a0
Commit Date:         None
Commit Branch:       None
Cargo Target Triple: x86_64-unknown-linux-gnu
Typst Version:       0.11.1
```

**Additional context**
https://github.com/Myriad-Dreamin/tinymist/releases/tag/v0.11.17
https://github.com/Myriad-Dreamin/tinymist/pull/365

Enter-tainer:
@kririae

kririae:
> @kririae

I'll work on this.

kririae:
I tried to fix the issue in https://github.com/Myriad-Dreamin/tinymist/pull/485.

Title completion is retained in `@` trigger for now, but it worth discussing whether or not it is the intended behavior.

quachpas:
Thanks!

PR #485, opened 2024-08-03
fix: apply label instead of bib title name in `at` completion

kririae:
When `at` is presented, open/close is neglected, thus the `apply` field is not correctly constructed.

A fix for https://github.com/Myriad-Dreamin/tinymist/issues/483


Myriad-Dreamin:
It deserves to add some tests to https://github.com/Myriad-Dreamin/tinymist/tree/0134de41c15428eecb84c21e6fe425ebf21e9ba8/crates/tinymist-query/src/fixtures/completion specific for the linked fixed issue.

PR #507, opened 2024-08-07
fix: use plural for 0 items

Nerixyz:
Previously, zero items would use the singular noun (e.g. "0 Word"). However, zero should also be followed by a plural (i.e. only 1 is followed by a singular). This PR fixes that.


Enter-tainer (approved):
nice catch

PR #525, opened 2024-08-11
feat: generate shell completion

Eric-Song-Nop:
I saw `clap_complete` is in Cargo.toml, but did not see it's usage, so I created a simple completion subcommand with it. Feel free to remove it if it's redundant work.

---

我看依赖里有`clap_complete`但是好像没有实现shell补全，就随便写了一个。假如其实已经有了就把这个issue 关了好了。


Myriad-Dreamin:
Thanks. the `clap_complete` dependency was copied from other project. I also copied the generation code from the project just now.

PR #526, opened 2024-08-12
feat: export text in range as ansi highlighted code

Myriad-Dreamin:


editors/vscode/package.json
```diff
@@ -801,6 +801,11 @@
         "category": "Typst",
         "icon": "$(file-pdf)"
       },
+      {
+        "command": "tinymist.copyAnsiHighlight",
+        "title": "Copy as Highlighted Code",
```
Enter-tainer, line 806: The title is a little bit misleading. I remember vscode already has such feature.  If you copy from vscode and paste to word, you can see highlighted code. I believe this is the way that most people think of "copy as highlighted".

Maybe consider rename this to "copy as ansi highlighted code" or something?

Myriad-Dreamin: "Copy as Ansi Code"? (we need short title otherwise it will make the context menu's width very fat)

PR #527, opened 2024-08-12
feat: find references for `Ref` and `Label`

Eric-Song-Nop:
image


Myriad-Dreamin:
I find two bugs in the "go to definition" of label.
+ Using the element's span as `range` and `name_range`, which is wrong.

+ The following code doesn't check whether the content is exact labelled by the querying label.

  https://github.com/Myriad-Dreamin/tinymist/blob/0fce237d579aebe233a9c80f9dfe3f6d9d1c0cd3/crates/tinymist-query/src/analysis/linked_def.rs#L101-L103

I have fixed the first bug, but not the second one. The second bug is visualized as following, query on the second label gives result that the definition is the first :

image

Possible fix, check the parent node and determine whether the element span is in the span of previous sibling of the label node, according to code: https://github.com/typst/typst/blob/3dbaf3a4ca3437489a8c08c1d459eb20d5f7466c/crates/typst/src/eval/markup.rs#L50-L59

crates/tinymist-query/src/analysis/linked_def.rs
```diff
@@ -100,19 +100,34 @@ pub fn find_definition(
...
                     let sel = Selector::Label(label);
                     let elem = introspector.query_first(&sel)?;
-                    let span = elem.span();
-                    let fid = span.id()?;
 
-                    let source = ctx.source_by_id(fid).ok()?;
-
-                    let rng = source.range(span)?;
+                    // if it is a label, we put the selection range to itself
+                    let (fid, rng, name_range) = if is_label {
```
Enter-tainer, line 105: nit:
i personally prefer `range`, `file_id` as variable name.

rng usuallly refers to random number generator

update: it looks like fid is actually a span id🤔

Myriad-Dreamin:
For the todo, we can find all files reachable during compilation and see the `ast::Ref` in these files.

https://github.com/Myriad-Dreamin/tinymist/blob/27aac0691611bdd06e0b68257dc35d520ea3a000/crates/tinymist-query/src/references.rs#L46-L48

The reachable files could be retrieved by `iter_dependencies`.

https://github.com/Myriad-Dreamin/tinymist/blob/0fce237d579aebe233a9c80f9dfe3f6d9d1c0cd3/crates/tinymist-query/src/symbol.rs#L41-L42

crates/tinymist-query/src/analysis/linked_def.rs
Myriad-Dreamin: `rng` is `range`, and `fid` is `file_id`.

Myriad-Dreamin: but there are 44 `fid` and 81 `rng` in the repository, if we are going to use longer name, we can fix them all once.

Myriad-Dreamin: > rng usuallly refers to random number generator

In the domain of program analysis, we are rarely use random number generator so there is no much misleading by using `rng` as `range`.

Eric-Song-Nop:
> I find two bugs in the "go to definition" of label.
> 
>     * Using the element's span as `range` and `name_range`, which is wrong.
> 
>     * The following code doesn't check whether the content is exact labelled by the querying label.
>       https://github.com/Myriad-Dreamin/tinymist/blob/0fce237d579aebe233a9c80f9dfe3f6d9d1c0cd3/crates/tinymist-query/src/analysis/linked_def.rs#L101-L103
> 
> 
> I have fixed the first bug, but not the second one. The second bug is visualized as following, query on the second label gives result that the definition is the first :
> 
> image

我感觉我不是完全理解第二个bug,看图的话是说有多个`definition for label`的时候会找到错误的那个吗？但是`label`应该是不允许重复定义的，哪怕是bib和label也不能是同名的，应该没有问题吧。

---

I am not so sure if I totally understand the 2nd bug. I understand from the image is that something will go wrong when there are multiple definitions for a label, but it should be forbidden by the spec?

crates/tinymist-query/src/analysis/linked_def.rs
Enter-tainer: > but there are 44 `fid` and 81 `rng` in the repository, if we are going to use longer name, we can fix them all once.

it's okay to ignore this comment if it is already widely used

Myriad-Dreamin:
> I am not so sure if I totally understand the 2nd bug. I understand from the image is that something will go wrong when there are multiple definitions for a label, but it should be forbidden by the spec?

https://github.com/user-attachments/assets/516bcf8a-a2b2-4f3a-acbc-4a5f4840f8f8

Eric-Song-Nop:
I got this, but it seems referencing them is still not allowed, right?
image

Myriad-Dreamin:
uh, I fixed the first bug since it is related. The client will issue a "goto definition" request first, if the definition covers the cursor, it will issue a "find references" request intermediately.

image

But as you imagine, I think the second bug in the linked definitions isn't quite related to this PR, and it is not quite straightforward to fix. So we can only focus the cross document reference ref todo in the code and leave the second bug for future.

---

> I got this, but it seems referencing them is still not allowed, right?

Correct.

> I might not be so familiar with typst query system, but this kind of aliased labels can only be queried? Or can they?

They can only be queried but not referenced.

Eric-Song-Nop:
I am trying to do workspace reference for Ref, but it fails on get source from path for non-root files, may u check it a little for me?

Eric-Song-Nop:
十分好用，希望合并，帅！

PR #575, opened 2024-08-27
feat: add gitpod layer

### LINKED ISSUE #382: Typst Preview fails on VSCode web and others.
**Describe the bug**
Typst Preview uses the address of the local host to communicate with the server for preview, so VSCode used from a web browser, such as Gitpod, VSCode web or Coder, will cause address mismatch and the preview will fail.

**Package/Software version:**

VSCode version(Help -> About):
```plain
Gitpod Code
Version: 1.90.2
Commit: dc0e7217f8f11d3249dcd2cd1d2152a2fdb45f3e
Date: 2024-06-20T12:06:32.291Z (2 wks ago)
Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:127.0) Gecko/20100101 Firefox/127.0
```

tinymist extension version: `v0.11.0`. Get it by `tinymist --version` in terminal.
```plain
Typst Version:       0.11.1
```

**Additional context**

https://github.com/Enter-tainer/typst-preview/blob/971dbc7f63ba12af4497204e7c433e0ed62d89a4/addons/frontend/src/main.js#L40

https://github.com/Enter-tainer/typst-preview/issues/292

tani:
Closes #382

Hi, thank you for this fantastic project.

Following the resolution of issue #382, I continued to investigate Tinymist on Gitpod. During my investigation, I discovered that Gitpod exposes local ports as subdomains of the workspace,
 for example, `www.example.com:8000` is mapped to `8000-www.example.com`.

This pull request adds a translation layer for the Typst preview URLs to accommodate this behavior.

### DIFF
--- a/editors/vscode/src/gitpod.ts
+++ b/editors/vscode/src/gitpod.ts
@@ -1,0 +1,33 @@
+/**
+ * Check if the current environment is Gitpod.
+ * @return True if the current environment is Gitpod, false otherwise.
+ */
+export function isGitpod(): boolean {
+  return !!process.env.GITPOD_WORKSPACE_ID && !!process.env.GITPOD_WORKSPACE_CLUSTER_HOST;
+}
+
+/**
+ * Create a Gitpod URL for the given URL string.
+ * @param urlstr The URL string to create a Gitpod URL for.
+ * @return The Gitpod URL
+ */
+export function translateGitpodURL(urlstr: string): string {
+  const url = new URL(urlstr);
+  if (!url.port) {
+    throw new Error("port is not specified in the URL");
+  }
+  if (!isGitpod()) {
+    throw new Error("Not in Gitpod environment");
+  }
+  if (url.protocol.match("ws")) {
+    url.protocol = "wss";
+  }
+  url.hostname =
+    url.port.toString() +
+    "-" +
+    process.env.GITPOD_WORKSPACE_ID +
+    "." +
+    process.env.GITPOD_WORKSPACE_CLUSTER_HOST;
+  url.port = "";
+  return url.toString();
+}
--- a/editors/vscode/src/gitpod.ts
+++ b/editors/vscode/src/gitpod.ts
@@ -17,7 +17,7 @@ export function translateGitpodURL(urlstr: string): string {
     throw new Error("port is not specified in the URL");
   }
   if (!isGitpod()) {
-    throw new Error("Not in Gitpod environment");
+    throw new Error("not in Gitpod environment");
   }
   if (url.protocol.match("ws")) {
     url.protocol = "wss";
--- a/editors/vscode/src/preview.ts
+++ b/editors/vscode/src/preview.ts
@@ -19,6 +19,15 @@ import {
   commandStartPreview,
   registerPreviewTaskDispose,
 } from "./extension";
+import { isGitpod, translateGitpodURL } from "./gitpod";
+
+function translateExternalURL(urlstr: string): string {
+  if (isGitpod()) {
+    return translateGitpodURL(urlstr);
+  } else {
+    return urlstr;
+  }
+}
 
 export function previewPreload(context: vscode.ExtensionContext) {
   getPreviewHtmlCompat(context);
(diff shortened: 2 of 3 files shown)


editors/vscode/src/preview-compat.ts
```diff
@@ -14,7 +14,18 @@ import {
 } from "./preview";
 import { tinymist } from "./lsp";
 import { loadHTMLFile } from "./util";
+import {
```
Myriad-Dreamin, line 17: I don't want to add new feature in `preview-compat.ts`, which may cause problem. People who want to use new features should move to tinymist.

Myriad-Dreamin (commented):
It looks good, but I would like to change the rust-side interface.

editors/vscode/src/preview-compat.ts
tani: Reverted. Thank you for your review.

crates/typst-preview/src/lib.rs
```diff
@@ -246,11 +246,15 @@ async fn preview_<T: CompileHost + Send + Sync + 'static>(
...
         })
     };
+    
     let data_plane_port = data_plane_port_rx.await.unwrap();
-    let html = html.replace(
-        "ws://127.0.0.1:23625",
-        format!("ws://127.0.0.1:{data_plane_port}").as_str(),
-    );
+    let url = format!("ws://127.0.0.1:{data_plane_port}");
+    let new_url = if gitpod::is_gitpod() {
```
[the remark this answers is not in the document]
tani: I solved your request. After merging #577, I will update my code to fit your code.

Myriad-Dreamin:
since https://github.com/Myriad-Dreamin/tinymist/pull/577, a preview server:
- requesting `http:/@GET` will respond the preview html
- requesting `ws:/` will establish a websocket connection to data plane.

This means that if gitpod serves your html with a proxying url, like `50000-workspace.gitpod.com`, the frontend html will try to connect the data plane by a relative url `/`, i.e. `50000-workspace.gitpod.com` itself as well. If that works, we may not be necessary to touch rust-side code anymore. The thing I'm not sure is whether it is connecting preview frontend via `50000-workspace.gitpod.com`. If it isn't, I would rather to merge your PR first so to avoid rebasing onto my changes.

See, since that PR, replacing data plane url with `/` will work:

https://github.com/Myriad-Dreamin/tinymist/blob/d35b869d4d18e45f0801bd3bb1d9b884958924ce/crates/tinymist/src/tool/preview.rs#L327-L328

https://github.com/Myriad-Dreamin/tinymist/blob/d35b869d4d18e45f0801bd3bb1d9b884958924ce/crates/typst-preview/src/lib.rs#L36-L47

Myriad-Dreamin:
image

tani:
With your excellent suggestion, the rust side code is already down to a few lines, and with the pull request changes you are working on now, I think that changes to the rust side code will probably not be necessary at all. I agree with your comment. However, the only concern is the SSL part of the protocol.

> Similar to issue https://github.com/Myriad-Dreamin/tinymist/pull/575, the frontend sometime will be served using https:// instead of http://. Therefore, ws:// will not work, and it should be changed to wss://. Could you provide a snippet that aligns with the security requirements of this protocol change?

Myriad-Dreamin:
> However, the only concern is the SSL part of the protocol.

I'm afraid this only applicable to the cases like running this extension behind remote instances, like gitpod. They will add a SSL proxy for us.
- for previewing in local, `wss://127.0.0.1:port` should not work as we surely don't have trusted SSL certificate for `127.0.0.1`.
- for previewing from gitpod, I have seen you already rewrote `ws` to `wss` in `translateGitpodURL`.

Is there other cases I'm missing?

tani:
First of all, your point is correct: wss does not work on localhost (127.0.0.1). I agree.

The case I should be telling you about now is the following conditional branch 

- For http use ws 
- For https use wss

This behaviour is expected, because using ws for https communication will result in an error output. (Modern browsers nowadays force the use of wss under https communication).

https://stackoverflow.com/questions/37712224/mixed-content-error-when-accessing-websocket-server-hosted-in-a-different-port

Myriad-Dreamin:
#577 has landed. I think we can just change typescript (extension) code for gitpod.

tani:
Noted. It's good to know that. I will work on this tonight!

tani:
Yes! I've got it working with minimal changes. Elegant!
Screenshot 2024-09-06 at 19 18 31

editors/vscode/src/gitpod.ts
```diff
@@ -0,0 +1,33 @@
...
+ * @param urlstr The URL string to create a Gitpod URL for.
+ * @return The Gitpod URL
+ */
+export function translateGitpodURL(urlstr: string): string {
+  const url = new URL(urlstr);
+  if (!url.port) {
+    throw new Error("port is not specified in the URL");
+  }
+  if (!isGitpod()) {
+    throw new Error("Not in Gitpod environment");
```
Myriad-Dreamin, line 20: `Not` -> `not`

Myriad-Dreamin:
> Yes! I've got it working with minimal changes. Elegant!

And obviously there is no bug introduced 😁(hopefully).

editors/vscode/src/gitpod.ts
tani: Yes!

PR #577, opened 2024-08-28
feat: listen data plane socket and serve frontend html on same address

Myriad-Dreamin:
- typst-preview now doesn't have any websocket dep now. instead it serves data plane socket on abstract message stream.
  ```rs
  type Message = Either<String, Bytes>;
  pub fn start_data_plane(
        &mut self,
        mut streams: mpsc::UnboundedReceiver<Future<Output = impl Sink<Message> + Stream<Message>>>,
    ) { ... }
  ```
- moved address and port arguments to `tinymist::tool::preview`, which do all concrete implementation inside.
- replacing data plane address with "/" directly.
  ```rs
  html.replace("ws://127.0.0.1:23625", "/")
  ```
todo:
- [x] backward combability


Enter-tainer:
quick question: why message is Either<String, Bytes> rather than Bytes?

crates/tinymist/src/tool/preview.rs
```diff
@@ -349,61 +371,134 @@ impl PreviewState {
...
-                            .unwrap();
-                        Ok(res)
-                    }
+        let websocket_tx = websocket_tx.clone();
+        service_fn(move |mut req: hyper::Request<Incoming>| {
+            let frontend_html = frontend_html.clone();
+            let websocket_tx = websocket_tx.clone();
+            async move {
+                // Check if the request is a websocket upgrade request.
+                if hyper_tungstenite::is_upgrade_request(&req) {
```
Enter-tainer, line 403: cool

crates/tinymist/src/tool/preview.rs
```diff
@@ -349,61 +371,134 @@ impl PreviewState {
...
             }
-            _ = rx => {
-                final_tx.send(()).ok();
+        }
+
+        tokio::select! {
+            _ = graceful.shutdown() => {
+                log::info!("Gracefully shutdown!");
+            },
+            _ = tokio::time::sleep(reflexo::time::Duration::from_secs(10)) => {
```
Enter-tainer, line 486: reflexo has its own time? interesting

Enter-tainer (approved):

crates/tinymist/src/tool/preview.rs
Myriad-Dreamin: it is an alias of https://docs.rs/crate/instant/latest which will work in browser as well. Not a real bonus, but we may be able to compile typst-preview to wasm now, since all system dependencies are already taken from the preview crate.

tani:
Similar to issue #575, the frontend sometime will be served using `https://` instead of `http://`. Therefore, `ws://` will not work, and it should be changed to `wss://`. Could you provide a snippet that aligns with the security requirements of this protocol change?

tani:
Is it possible to add the following snippet?

```diff
/// Placeholders for typst-preview program initializing frontend
/// arguments.
function retrieveWsArgs() {
    /// The string `preview-arg:previewMode:Doc` is a placeholder
    /// It will be replaced by the actual preview mode.
    /// ```rs
    ///   let frontend_html = frontend_html.replace(
    ///     "preview-arg:previewMode:Doc", ...);
    /// ```
    let mode = 'preview-arg:previewMode:Doc';
    /// Remove the placeholder prefix.
    mode = mode.replace('preview-arg:previewMode:', '');
    let previewMode = PreviewMode[mode];

    /// The string `ws://127.0.0.1:23625` is a placeholder
    /// Also, it is the default url to connect to.
    let url = "ws://127.0.0.1:23625";
+    if (location.href.startsWith("https://")) {
+        url = url.replace("ws://", "wss://")
+    }
    /// Return a `WsArgs` object.
    return { url, previewMode, isContentPreview: false };
}
```

Cargo.lock
```diff
@@ -1510,7 +1579,7 @@ checksum = "ec3efd23720e2049821a693cbc7e65ea87c72f1c58ff2f9522ff332b1491e590"
 dependencies = [
  "futures-util",
  "http 0.2.12",
- "hyper",
```
Enter-tainer, line 1513: unfortunately we have both hyper 1.0 and hyper 0.14 now

Myriad-Dreamin: This will not be a problem after we upgrading reqwest to latest.

tani:
@Myriad-Dreamin Sure. I will create the PR.

PR #582, opened 2024-08-31
refactor: add compatibility for `typst_syntax::LinkedNode.leaf_at`

ParaN3xus:
Add compatibility for `typst_syntax::LinkedNode.leaf_at`

### DIFF
--- a/crates/tinymist-query/src/analysis.rs
+++ b/crates/tinymist-query/src/analysis.rs
@@ -97,7 +97,7 @@ mod post_type_check_tests {
 
     use insta::with_settings;
     use typst::syntax::LinkedNode;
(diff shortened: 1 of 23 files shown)


Enter-tainer (approved):
LGTM

Myriad-Dreamin:
Please don't use macro but use function instead whenever possible, because macros are not safely typed. For other code, also LGTM.

Myriad-Dreamin:
Note we can apply some tricks to do better than provide functions. we can implement a `LinkedNodeExt` trait.

```rs
trait LinkedNodeExt {
  fn leaf_at_compat(&self, ...);
}
```

and usage:
```rs
use LinkedNodeExt;
let node = root.leaf_at_compat(pos + 1).unwrap();
```

PR #598, opened 2024-09-13
feat: show first occurrence locations for used fonts

hooyuser:
This PR adds a new feature to the summary command. It now includes the ability to display the locations of the first occurrences of used fonts within Typst source files. Two new fields are added:

- `first_occur_file`: the file path of the Typst source containing the locatable text element where the font first occurs.

- `first_occur_line`: The line number of the locatable text element where the font first occurs.

- `first_occur_column`: The column number of the locatable text element where the font first occurs.


crates/tinymist-query/src/document_metrics.rs
```diff
@@ -149,12 +163,49 @@ impl<'a, 'w> DocumentMetricsWorker<'a, 'w> {
...
+        let glyph_len = text.glyphs.len();
+
+        let has_source_info = if let Some(font_info) = self.font_info.get(&font_key) {
+            font_info.first_occur_file.is_some()
+        } else {
+            false
+        };
+
+        if !has_source_info && glyph_len > 0 {
+            let (span, span_offset) = text.glyphs[0].span;
```
Myriad-Dreamin, line 176: You seems to be ignoring `span_offset`,  this could be a large value (at most 65535 offset from the `span`) and cause imprecise reporting

Myriad-Dreamin (commented):
It looks well other than a possible missing. I left a comment there.

crates/tinymist-query/src/document_metrics.rs
hooyuser: I used the same logic as in this code: https://github.com/Myriad-Dreamin/tinymist/blob/58fea291ad8ffe942b51c52bd9225a6d344a01a5/crates/tinymist/src/main.rs#L250-L257 I haven’t looked deeply into what `span_offset` exactly refers to, but based on some quick experiments, it seems like `text.glyphs[i].span` just returns `{span_offset: i}`. If we only focus on `glyphs[0]`, I guess it should work?

hooyuser: When getting the span, the info that it comes from the i-th glyph gets lost. In this case, I think we should add an offset to fix it, like here:
https://github.com/Myriad-Dreamin/tinymist/blob/58fea291ad8ffe942b51c52bd9225a6d344a01a5/crates/tinymist/src/tool/preview.rs#L85-L107 
I can add similar logic, but I’m not sure if it's really necessary.
```rust
let range = source.range(span)?;
let byte_index = if offset < range.len() as u16 {
    range.start + offset as usize
};
let line = source.byte_to_line(byte_index)?;
```

Myriad-Dreamin: > it seems like text.glyphs[i].span just returns {span_offset: i}

You could try checking `"测试"`, it is probably the (utf-8) byte offset related to the span's start position.

Myriad-Dreamin: > If we only focus on glyphs[0]

This only works occasionally since the first glyph of a text can be relative to some span with non-zero offset.

hooyuser: Thank you! I’ve fixed it. You were right about the span's byte offset.

Myriad-Dreamin:
Lets improve UI (if necessary) in a separated PR.

PR #636, opened 2024-10-06
feat: add configuration to open exported file by system default app

Myriad-Dreamin:


memeplex:
Hi! I'm not seeing this as an enumeration but as an editable field for which it's almost impossible to figure out the allowed values without looking at the source.

image

image

image

Issue #534, opened 2024-08-15
Clicking on preview opens a new window when working with multiple typst files


ziiirozone:
**Describe the bug**
Hello !
I'm currently working with vs code on two different typst files, which I both have previews of. For the sake of explainations I'll number them file 1, file 2 & preview 1, preview 2. When I interact with file 1 and click on a word in preview 1 everything works as expected, but when I click on a word in preview 2, a new window with file 2 opens with the cursor on the word I clicked on. I now have 2 windows with file 2 open. When I double click on a word of preview 2 it seems to remember the first instance of file 2 and goes back to it. This makes working on multiple typst files quite difficult.

VSCode version :
```plain
Version: 1.92.1 (user setup)
Commit: eaa41d57266683296de7d118f574d0c2652e1fc4
Date: 2024-08-07T20:16:39.455Z
Electron: 30.1.2
ElectronBuildId: 9870757
Chromium: 124.0.6367.243
Node.js: 20.14.0
V8: 12.4.254.20-electron.0
OS: Windows_NT x64 10.0.22631
```

tinymist version 0.11.19

Myriad-Dreamin:
The opening to new window function usually works well when the layout simply looks like following:

![image](https://github.com/user-attachments/assets/cdb69410-0054-4b7d-a095-94cd87707361)

But thing goes worse rapidly in other layouts, and we don't know how many layouts we should adapt to. I believe this is likely a twist to tune for previewing, and it is not quite related to tinymist's core function. The current algorithm is used here:

https://github.com/Myriad-Dreamin/tinymist/blob/68911d91cbaea2e9fb91f33df3d803b5b433ee41/editors/vscode/src/preview.ts#L684

https://github.com/Myriad-Dreamin/tinymist/blob/68911d91cbaea2e9fb91f33df3d803b5b433ee41/editors/vscode/src/util.ts#L24-L33

ziiirozone:
I have no knowledge on how vs. code works or even how to program in typescript, but wouldn't it be possible to loop over each editor until one is found with a title matching that of the preview ?

memeplex:
I'm seeing this quite frequently too. I'm trying to find some recipe in order to reproduce it in a more reliable fashion, still to no avail. By now I do can say that:

- Sometimes it happens even in the basic layout described above, but it might be that this is after having opened another previews (and then closed them), or because of having other files at the left group, or some of them being typ files, IDK, I can say it happens quite often but I still haven't found the pattern.

[...]

PR #734, opened 2024-10-26
feat: add preview icon when clicking outside the document

### LINKED ISSUE #615: Preview icon disappears when clicking outside the document
This issue came from [Typst Preview#262](https://github.com/Enter-tainer/typst-preview/issues/262)

**Describe the bug**
When I click outside the typst document (but keeping it open), the preview icon in the top right-hand corner disappears.

https://github.com/Enter-tainer/typst-preview/assets/54775010/a8a30463-3d27-4fc6-b3a7-c253a08b2947

**To Reproduce**
+ Open a typst document
+ Click outside the document (in the terminal or in the directory tree)

**Expected behavior**
The preview icon should remain visible as long as the document is open, even if the user clicks outside the document.

**Package/Software version:**

```plain
tinymist
Build Timestamp:     2024-08-26T16:22:10.986712173Z
Build Git Describe:  v0.11.20
Commit SHA:          1f9a42c2e863f6b1fe8624da796e47adf44a9a83
Commit Date:         None
Commit Branch:       None
Cargo Target Triple: x86_64-unknown-linux-gnu
Typst Version:       0.11.1
```

supersurviveur:
Here is another try to fix #615.

I store the current activeEditor to retreive it even if the user is in the output pannel. I change the activeEditor when another typst document is opened.
This way the preview button always works, even in the output pannel.


Myriad-Dreamin:
Hi~ Thank you for go on fixing this issue. The current solution still has some problems:
- Both `vscode.TextEditor` and `vscode.TextDcoument` can be disposed at any time, so we cannot access or hold them across callbacks. It also prevents GC of these objects.
- If all typst documents are not visible in window, we should hide the preview icon, to save precious space in the top bar. To control visibility programatically, we can use [contextKey](https://code.visualstudio.com/api/references/when-clause-contexts#add-a-custom-when-clause-context).

  Example usage:

  https://github.com/Myriad-Dreamin/tinymist/blob/2c38695b6f922d9978d2de53c36c534a17cf0a58/editors/vscode/src/extension.ts#L65-L67

  https://github.com/Myriad-Dreamin/tinymist/blob/2c38695b6f922d9978d2de53c36c534a17cf0a58/editors/vscode/package.json#L47

supersurviveur:
The second issue seems to be easy to solve with a context. But I think that it's not even needed, because the preview icon is shown only if a typst document is opened and is visible.

The first one is harder to solve, maybe we can check is the `textEditor` is disposed when the preview is launched, and use the `window.activeEditor` by default.

supersurviveur:
Now the activeEditor is dropped when it is closed. Moreover, `window.activeEditor` is used if the editor stored is `undefined`. This way the second issue is not a problem and it's less a problem for the GC.

However, now when I click on the document, it shows me the source in another textEditor...

Myriad-Dreamin:
> However, now when I click on the document, it shows me the source in another textEditor...
> EDIT: I also have the issue without my modifications.

https://github.com/Myriad-Dreamin/tinymist/issues/534

Issue #778, opened 2024-11-08
Add unicode input by symbol expansion


owiecc:
**Motivation**

It would be nice to be able to use the typst symbol input and be able to expand it to the corresponding glyph. This way equations like this: `$ nabla dot.c E = rho / epsilon_0 $` can be converted to a simpler notation: `$ ∇·E = ρ / ε_0 $`. Both produce the same output but the latter is far more readable.

**Description**

Typing `dot.c<TAB>` converts the `dot.c` to `·` in the source document. 

**Examples/Questions**

This is a similar idea to the way one can input symbols in [Julia](https://docs.julialang.org/en/v1/manual/unicode-input/) with <TAB> completion.

I have proposed this feature to the, now deprecated, typst-lsp: https://github.com/nvarner/typst-lsp/issues/49 previously. 

Relevant discussion on Julia discourse: https://discourse.julialang.org/t/unicode-autocomplete-howto/96752/2

Relevant typst extension: https://github.com/supersurviveur/typst-math

I am willing to set up a bounty for this feature.

owiecc:
I think starting with manual expansion with a `<TAB>` and testing the possibility of automating it is the easiest way to implement it. Manual expansion has none of the drawback that you mentioned and people can choose when to use it. Nothing changes for current users, they will not even notice the change. Once implemented the auto-expansion can be implemented on top of it. 

From my experience with Julia, the `<TAB>` is not that problematic. Adding one keystroke is not a lot when most symbol names in Typst have few to dozen characters. Remember that not all symbols need to be expanded. Maybe it is beneficial to expand only a few characters in key places to increase readability of an equation.

Enter-tainer:
> Maybe it is beneficial to expand only a few characters in key places to increase readability of an equation.

this might be a good idea. we may firstly adopt symbol list from typst-math extension

Eric-Song-Nop:
I am not sure what are u really looking for? Is it more like an input method for unicode or an immediate renderer for a subset of math mode? If the first solution, as an ime, is what u need, it is a common pattern to implement this mode with an extra trigger. I believe in Julia or Lean or Agda extension, start with `\` triggers such an ime.

Eric-Song-Nop:
BTW, I also made a nvim-cmp source for this purpose: https://github.com/Eric-Song-Nop/cmp-typst-symbols

PR #788, opened 2024-11-10
basic unicode ime

Eric-Song-Nop:
A basic implementation for #778, a Unicode IME for typst symbols.

image
As demonstrated in the image, the `\alpha` completion will be expanded as the alpha symbol instead of `alpha` text.

TODO:
- [ ] Discuss the necessity of this feature
- [ ] How this should be triggered
- [ ] Symbols with alternatives


Myriad-Dreamin:
Using `\` to trigger unicode ime looks a bit offensive /s 😄

Back to PR, I think it is not great to use `\` to trigger this function because we have already use `\` for escaping characters.

Eric-Song-Nop:
I am mostly concerned about:
1. Should this feature be included in the language server?
    Some servers I know does have this feature, mostly languages design for math or linear algebra. While I always take it as some feature that other plugins can provide, like this one: https://github.com/Eric-Song-Nop/cmp-typst-symbols, while it might also be nice to be included in language server so that the symbol table can be updated with typst, and also since tinymist is a full toolbox for typst, it might also be nice to have it.
 2. The trigger letter. To me, any trigger letter is fine, while Typst has used some symbols, I can think of a better one, maybe `%`?
 3. Should it be triggered only in math mode? if so, we might not need this feature at all.
 4. Use we include emojis? There might be too many emojis included by typst.

Myriad-Dreamin:
> Should this feature be included in the language server?

If we would like to expand all valid symbol in scope, we should include it in the language server. For example, we will fail to expand a symbol aliased by some let soonly if we cannot check scopes, which requires code analysis:

```typ
#let grad = math.nabla;
$ \grad $ // it fails to expand by `\`
```

> The trigger letter.

I personally think we should not use special trigger characters, but we could use commit character. E.g., when some user presses a `Tab` key, the editor will commit the completion item expanding the symbol. This is doable as mentioned in the issue #778.

> Should it be triggered only in math mode? if so, we might not need this feature at all.

I don't quite understand "if so, we might not need this feature at all." But we don't have reason to not doing so in markup or code mode.

> Use we include emojis?

It is not a problem if we include this feature in the language server. We get the table of emojis from the builtin modules.

Myriad-Dreamin:
There are too many problems to solve so we should close this hanging PR. We can summarize the design issues and failures in the origin issue. Please feel free to reopen it or open another PR.

Issue #799, opened 2024-11-12
Document formatting unavailable in editor


MrLego8-9:
**Describe the bug**

The formatting option is unavailable in Kate and Helix.

While Kate is not officially supported by tinymist, it implements the LSP protocol.

Here are the LSP settings in Kate (they're the same in Helix)

```json
"settings": {
				"exportPdf": "onSave",
				"formatterMode": "typstyle"
			}
```

From what I saw in other issues, it may be due to dynamic registration.

Is there any way to fix this ?

**Package/Software version:**
(From arch linux repos)
tinymist extension version: `v0.12.0`. Get it by `tinymist --version` in terminal.
```plain
tinymist
Build Timestamp:     2024-11-11T13:20:38.000000000Z
Build Git Describe:  VERGEN_IDEMPOTENT_OUTPUT
Commit SHA:          VERGEN_IDEMPOTENT_OUTPUT
Commit Date:         None
Commit Branch:       None
Cargo Target Triple: x86_64-unknown-linux-gnu
Typst Version:       0.12.0
Typst Source:        git+https://github.com/Myriad-Dreamin/typst.git?tag=tinymist-v0.12.0#58426a90a7ef721738a01be09793d33e55eb75a9
```

Eric-Song-Nop:
I am not familiar with Kate, while the formatterMode setting works on my side on helix:
```toml
[language-server.tinymist]
command = "tinymist"
config = { exportPdf = "onType", formatterMode = "typstyle" }

[[language]]
name = "typst"
language-servers = ["tinymist"]
```

It would be nice if you can provide us with your lsp.log file from the editor.

MrLego8-9:
[kate_lsp.log](https://github.com/user-attachments/files/17733862/kate_lsp.log)
[tinymist.log](https://github.com/user-attachments/files/17733866/tinymist.log)
Here are both logs, from Kate and Tinymist.

By the way, formatting with LSP works in Kate, as it works out of the box with the Haskell LSP.

Myriad-Dreamin:
I notice that kate doesn't send configuration when initializing server, which prevents server from enabling formatter, because tinymist will enable in exact following cases:
- when formatter mode is set in configuration in the initialization params.
- when dynamic registration is supported and formatter mode configuration is changed.

kate doesn't trigger all of above cases.

Myriad-Dreamin:
Fwiw, I was hesitating to remove following code, inherited from typst-lsp:

[...]

PR #814, opened 2024-11-14
fix: statically register formatting and semantic tokens capabilities

MrLego8-9:
A fix for #799.
Instead of dynamically registering formatting and semantic tokens capabilities, register them statically.

The toggles in the config works the same, for formatting and semantic highlighting. This, however, improves compatibility with some editors that do not support dynamic registering in their LSP client implementation.

I have tested the changes in VSCode (CodeOSS), Helix and Kate (does not support dynamic registration).

### DIFF
--- a/tests/e2e/main.rs
+++ b/tests/e2e/main.rs
@@ -374,7 +374,7 @@ fn e2e() {
         });
 
         let hash = replay_log(&tinymist_binary, &root.join("neovim"));
-        insta::assert_snapshot!(hash, @"siphash128_13:37db2c66ca1b5a0cdec20f531c15d5");
+        insta::assert_snapshot!(hash, @"siphash128_13:fbb7e6bec2bd8b4d14f29b9d36066d20");
     }
 
     {


Myriad-Dreamin:
Could you inline these two variable? The comment above is not needed as well.

crates/tinymist/src/init.rs
```diff
@@ -150,22 +150,15 @@ impl Initializer for SuperInit {
...
-        };
-        let document_formatting_provider = match service.config.formatter_mode {
-            FormatterMode::Typstyle | FormatterMode::Typstfmt
-                if !const_config.doc_fmt_dynamic_registration =>
-            {
-                Some(OneOf::Left(true))
-            }
-            _ => None,
-        };
+        let semantic_tokens_provider = Some(
```
Myriad-Dreamin, line 153: Could you inline these two variable? The comment above is not needed as well.

MrLego8-9: Fixed

Myriad-Dreamin:
I have tested it on my vscode, it doesn't not work if we always register them statically, so I turn to register them only if dynamic registering is not supported. Since we have made a mistake, I might delay to merge it in v0.12.2+ to reduce risk to break things.

MrLego8-9:
> I have tested it on my vscode, it doesn't not work if we always register them statically, so I turn to register them only if dynamic registering is not supported. Since we have made a mistake, I might delay to merge it in v0.12.2+ to reduce risk to break things.

Sure, no problem, if you think it's better to wait in order not to break things, go ahead.

Issue #852, opened 2024-11-19
abnormal auto completion got from VS Code tinymist extension


mmonci:
**Describe the bug**
the hint of image file path always contain **an unexpected ”../“**. Got it every time I call image function.

An example shown below.

![show](https://github.com/user-attachments/assets/e2573211-20b1-4e12-ad68-cf4346af6aa0)

**Package/Software version:**

tinymist extension version: `v0.12.2`. I just downloaded the extension in VS Code.

Eric-Song-Nop:
I can confirm that it is also happening on Neovim, so it's a bug on the server side.

I am looking into this bug

PR #854, opened 2024-11-19
fix: wrong diff_path

Eric-Song-Nop:
A fix for #852

In short the crate pathdiff::diff_paths is used incorrectly. Here is an example:
```rust
assert_eq!(diff_paths("/foo/bar/quux", "/foo/bar/baz"),  Some("../quux".into())); // thus the following
assert_eq!(diff_paths("/foo/bar/image.jpg", "/foo/bar/file.typ"),  Some("../image.jpg".into()));
```
We might add more test to inspect other cases

### DIFF
--- a/crates/tinymist-query/src/upstream/complete/ext.rs
+++ b/crates/tinymist-query/src/upstream/complete/ext.rs
@@ -1484,10 +1484,17 @@ pub fn complete_path(
             // diff with root
             unix_slash(path.vpath().as_rooted_path())
         } else {
-            let base = base.vpath().as_rooted_path();
-            let path = path.vpath().as_rooted_path();
-            let w = pathdiff::diff_paths(path, base)?;
-            unix_slash(&w)
+            let base = base.vpath().as_rooted_path().parent();
+            let path = path.vpath();
+            match base {
+                Some(base) => {
+                    let w = pathdiff::diff_paths(path.as_rooted_path(), base)?;
+                    unix_slash(&w)
+                }
+                None => {
+                    unix_slash(path.as_rootless_path())
+                }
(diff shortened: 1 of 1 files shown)


Myriad-Dreamin:
Nice!

> We might add more test to inspect other cases

We should add necessary fixtures for essential completion features, but we only set up a part of them. To add fixtures to cover cases of completing import paths, we could see:
- fixture containing multiple files: https://github.com/Myriad-Dreamin/tinymist/blob/main/crates/tinymist-query/src/fixtures/completion/completion_title.typ
- fixture testing completion inside strings: https://github.com/Myriad-Dreamin/tinymist/blob/main/crates/tinymist-query/src/fixtures/completion/modify_string.typ

crates/tinymist-query/src/upstream/complete/ext.rs
```diff
@@ -1484,10 +1484,17 @@ pub fn complete_path(
...
-            let w = pathdiff::diff_paths(path, base)?;
-            unix_slash(&w)
+            let base = base.vpath().as_rooted_path().parent();
+            let path = path.vpath();
+            match base {
+                Some(base) => {
+                    let w = pathdiff::diff_paths(path.as_rooted_path(), base)?;
+                    unix_slash(&w)
+                }
+                None => {
```
Myriad-Dreamin, line 1494: We could simplify code:

```rs
let base = base.vpath().as_rooted_path().parent().unwrap_or(Path::new("/"));
```

Eric-Song-Nop:
我可能现在没空写测试，要是着急的话可以先merge 我再开新的pr写测试

PR #885, opened 2024-11-22
fix: target to use for cross build and bump version to 0.12.4-rc3

Myriad-Dreamin:


.github/workflows/release.yml
```diff
@@ -41,35 +41,39 @@ jobs:
...
             regular_build: 'true'
             cross: true
           - os: ubuntu-20.04
-            rust-target: aarch64-unknown-linux-gnu
+            rust-target: aarch64-unknown-linux-musl
             platform: linux
             arch: arm64
             cross: true
           - os: ubuntu-20.04
-            rust-target: arm-unknown-linux-gnueabihf
```
Enter-tainer, line 60: not sure about this. i dont know the difference between arm targets

.github/workflows/release.yml
```diff
@@ -41,35 +41,39 @@ jobs:
...
             regular_build: 'true'
+            cross: false
           - os: windows-latest
             rust-target: aarch64-pc-windows-msvc
             platform: win32
             arch: arm64
+            cross: false
           - os: ubuntu-20.04
-            rust-target: x86_64-unknown-linux-gnu
+            rust-target: x86_64-unknown-linux-musl
```
Enter-tainer, line 51: this can help with rhel. but i heard that some libc function in musl is inefficient. we can observe if there is any slow down after changing this.

.github/workflows/release.yml
Myriad-Dreamin: We ran into some errors in https://github.com/Myriad-Dreamin/tinymist/pull/884. I guessed we should change targets supported by cross after diff typst CI scripting with ours.

.github/workflows/release.yml
```diff
@@ -41,35 +41,39 @@ jobs:
...
             platform: linux
             arch: x64
             regular_build: 'true'
             cross: true
```
Coekjan, line 55: Building `*-musl` might require `cross: true`, afaik.

Coekjan:
Testing the `*-musl` requires more configs, like `[target.x86_64-unknown-linux-musl.linker]` in `.cargo/config.toml`. So testing fails. See:

```diff
      - name: Test tinymist
        run: |
-          cargo test --profile=gh-release --workspace --target ${{ matrix.rust-target }}
+          cross test --profile=gh-release --workspace --target ${{ matrix.rust-target }}
```

Maybe we can just use `cross` for testing, and get rid of dealing with the complex configs...

Coekjan:
> https://github.com/Myriad-Dreamin/tinymist/actions/runs/11975350731/job/33388515141?pr=885

Seems that cross-testing `*-musl` targets would also require vendored openssl.

Coekjan:
Seems that `build_alpine` is similiar (same?) with `build linux-x64` now?

Myriad-Dreamin:
@Coekjan I've run `cross test ... --features vendor-openssl`, it now at least starts successfully, but it seems to be failing when downloading packages.

```
called `Result::unwrap()` on an `Err` value: MalformedArchive(Some("failed to create `/.cache/typst/packages/preview/tidy/0.3.0`"))
```

> Seems that build_alpine is similiar (same?) with build linux-x64 now?

Pretty same. Now they built the same target. The only difference is in that build alpine is built in an alpine image. I suspect whether we can remove it right now, but I would rather not touching them until release 0.12.4.

Myriad-Dreamin:
I'll close it and leave this issue for next release.

PR #894, opened 2024-11-24
feat: restart server command in vscode

Eric-Song-Nop:
The upstream bugs might crash the language server in many cases, I feel it a requirement to have an explicit way to restart the language server client in such case.

### DIFF
--- a/editors/vscode/package.json
+++ b/editors/vscode/package.json
@@ -1020,7 +1020,7 @@
         },
         {
           "command": "tinymist.restartServer",
-          "when": "editorLangId == typst"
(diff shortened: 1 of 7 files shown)


Myriad-Dreamin:
Some thoughts:
- We could detect and restart the server once after user editing typst document. It is not done intermediately after server down, to avoid infinite or aggressively restart.
- The right-below status bar could hints the user that currently the server is down for some reason, e.g. "Crashed", hovering with message "Compiler panicked: blabla".

Myriad-Dreamin (approved):

PR #922, opened 2024-12-02
feat: search template gallery

Eric-Song-Nop:
Basic search template gallery by name.

Will update with fuzzy finder, and search by description later.

Might also be nice to have search and insert universe for packages also.

### DIFF
--- a/tools/editor-tools/src/features/template-gallery.ts
+++ b/tools/editor-tools/src/features/template-gallery.ts
@@ -75,11 +75,11 @@ const TemplateList = (
     );
   };
 
-  const highlightMatch = (text: string, searchTerm?: string) => {
-    if (!searchTerm || !text) return van.tags.span({}, text);
-    console.log("searchTerm", searchTerm);
+  const highlightMatches = (text: string, searchResults?: SearchResult[]): HTMLSpanElement => {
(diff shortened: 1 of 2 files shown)


Myriad-Dreamin:
I reimplemented it by using minisearch and indexing fields: `["name", "description", "authors", "keywords", "categories"]`. The next missing feature is highlighting search terms in page. For example, when the input is "stu", the "students" word is highlighted in the description "All-in-one template for students and theses".

image

Eric-Song-Nop:
I have accomplished the highlight in desc and title, with a yellow bg. While, it seems that matches in keywords/authors are hard to highlight(maybe not needed for now

Myriad-Dreamin:
Wow, you made it like a flash.

> I have accomplished the highlight in desc and title, with a yellow bg. 

The yellow bg looks not 100% fit the dark themes, but should be already pretty nice.

> While, it seems that matches in keywords/authors are hard to highlight(maybe not needed for now

We might highlight the whole boxes for them. For example, with entering "pres" page will show the "presentation" keyword box with a different foreground and background color at the same time.

image

Myriad-Dreamin (approved):

PR #926, opened 2024-12-02
test(ci): add fmt, clippy, doc, and msrv tests to CI

Myriad-Dreamin:


Enter-tainer (approved):
i'm surprising to know fmt and clippy isn't set up before

Myriad-Dreamin:
> i'm surprising to know fmt and clippy isn't set up before

Copied from typst-preview

Issue #945, opened 2024-12-04
Import items no longer being autocompleted in v0.12.8


Andrew15-5:
**Describe the bug**

Yesterday and today, I noticed that the latest version doesn't show up any items in the completion dropdown menu (in Neovim v0.10.0) when trying to look into what a package has to offer. Pretty sure it's a regression.

```typ
#import "@preview/package:0.0.0":
```

Moreover, before this version I had a small issue where the items show up only if I request them after `: `, but it will not show them if I've already written something (e.g., `: package-ite`).

**Package/Software version:**

tinymist extension version: `v0.11.0`. Get it by `tinymist --version` in terminal.
```plain
tinymist
tinymist
Build Timestamp:     2024-12-03T14:31:25.526257462Z
Build Git Describe:  v0.12.8
Commit SHA:          27460a9c53960be8803dd0ff23e20fa9740987fd
Commit Date:         None
Commit Branch:       None
Cargo Target Triple: x86_64-unknown-linux-gnu
Typst Version:       0.12.0
Typst Source:        git+https://github.com/Myriad-Dreamin/typst.git?tag=tinymist-v0.12.0#58426a90a7ef721738a01be09793d33e55eb75a9
```

**Logs:**

tinymist server log(Output Panel -> tinymist):

<details><summary>logs</summary>

```plain
[START][2024-12-04 12:50:48] LSP logging initiated
[2024-12-04T09:50:48Z INFO  tinymist] tinymist version information: [["Build Timestamp", "2024-12-03T14:31:25.526257462Z"], ["Build Git Describe", "v0.12.8"], ["Commit SHA", "27460a9c53960be8803dd0ff23e20fa9740987fd"], ["Commit Date", "None"], ["Commit Branch", "None"], ["Cargo Target Triple", "x86_64-unknown-linux-gnu"], ["Typst Version", "0.12.0"], ["Typst Source", "git+https://github.com/Myriad-Dreamin/typst.git?tag=tinymist-v0.12.0#58426a90a7ef721738a01be09793d33e55eb75a9"]]
[2024-12-04T09:50:48Z INFO  tinymist] starting Language server: LspArgs {
        mirror: MirrorArgs {
            mirror: "~/tinymist-input-mirror.json",
            replay: "",
        },
        font: CompileFontArgs {
            font_paths: [],
            ignore_system_fonts: false,
        },
    }
[...]

PR #949, opened 2024-12-05
fix: Import Item Completion

Eric-Song-Nop:
Part fix for #945 

- Fix: for `import "lib.typ": tes|`, the parent syntax is import item path, then parent is import items

### DIFF
--- a/crates/tinymist-query/src/upstream/complete.rs
+++ b/crates/tinymist-query/src/upstream/complete.rs
@@ -548,6 +548,8 @@ fn complete_imports(ctx: &mut CompletionContext) -> bool {
     if_chain! {
         if ctx.leaf.kind() == SyntaxKind::Ident;
         if let Some(parent) = ctx.leaf.parent();
+        if parent.kind() == SyntaxKind::ImportItemPath;
+        if let Some(parent) = parent.parent();
         if parent.kind() == SyntaxKind::ImportItems;
         if let Some(grand) = parent.parent();
         if let Some(ast::Expr::Import(import)) = grand.get().cast();


Andrew15-5:
So, is this an all-editors fix or does import items completion currently work in some editors?

Myriad-Dreamin:
Looks like a patch that can be ported to the upstream, [here](https://github.com/typst/typst/blob/main/crates/typst-ide/src/complete.rs#L519)

Myriad-Dreamin:
It is not completely fixed, if we start to really use the `ImportPath` syntax, as the code [here](https://github.com/typst/typst/blob/79a7a6bf77037e0b6124193ebc8b3c5e8a0528bb/crates/typst-ide/src/complete.rs#L553) doesn't consider the newly introduced syntax as well.

Eric-Song-Nop:
> Looks like a patch that can be ported to the upstream, [here](https://github.com/typst/typst/blob/79a7a6bf77037e0b6124193ebc8b3c5e8a0528bb/crates/typst-ide/src/complete.rs#L519)

Yes, I tested in the webapp and there is such issue there. Let me do it, I haven't contributed to Typst yet :)

Myriad-Dreamin (approved):

Myriad-Dreamin:
In fact, there are 6 related bugs to fix at the same time.

Andrew15-5:
Why there is no LF at the end of the Typst files? This (1) triggers me and (2) adds unnecessary visual noise into diffs so they become harder to read.

Myriad-Dreamin:
> Why there is no LF at the end of the Typst files? This (1) triggers me and (2) adds unnecessary visual noise into diffs so they become harder to read.

@Andrew15-5 The only reason might be the formatter doesn't add LF at the end for me and I'm not used to adding them by myself. @Enter-tainer

Andrew15-5:
Yeah, I think in Neovim it automatically adds this + it hides the last empty line. But in GUI editors I don't know if there is a built-in feature for this. In the web app I just have to add it manually, like I did this before I switched to Neovim.

PR #959, opened 2024-12-07
fix: simplify types of variables for hover tooltip

Eric-Song-Nop:
better hover on var:
```typst
#let d = (:)
       ^
       |
       let d = any; => let d = dictionary;
```

### DIFF
--- a/crates/tinymist-query/src/hover.rs
+++ b/crates/tinymist-query/src/hover.rs
@@ -178,7 +178,21 @@ fn def_tooltip(
 
                         match &sym_docs {
                             Some(DefDocs::Variable(docs)) => {
-                                push_result_ty(def.name(), docs.return_ty.as_ref(), &mut type_doc);
+                                if let Some((_, long, _)) = docs.return_ty.as_ref() {
+                                    if long.as_str() == format!("@{}", def.name()) {
+                                        let ty = ctx
+                                            .literal_type_of_node(leaf)?
+                                            .repr()
+                                            .unwrap_or_else(|| "any".into());
+                                        let _ = write!(type_doc, " = {ty}");
+                                    } else {
+                                        push_result_ty(
+                                            def.name(),
+                                            docs.return_ty.as_ref(),
+                                            &mut type_doc,
+                                        );
+                                    }
+                                }
                             }
                             Some(DefDocs::Function(docs)) => {
                                 let _ = docs.print(&mut type_doc);
--- a/crates/tinymist-query/src/hover.rs
+++ b/crates/tinymist-query/src/hover.rs
@@ -178,21 +178,7 @@ fn def_tooltip(
 
                         match &sym_docs {
                             Some(DefDocs::Variable(docs)) => {
-                                if let Some((_, long, _)) = docs.return_ty.as_ref() {
-                                    if long.as_str() == format!("@{}", def.name()) {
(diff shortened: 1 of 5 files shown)


Myriad-Dreamin:
The root cause may be we are only parsing and returning docstring on variables, no type checking is performed during this process:

https://github.com/Myriad-Dreamin/tinymist/blob/5dd9badc112c1a8f864368261ef5b9692ac621bb/crates/tinymist-query/src/analysis/global.rs#L462-L465

Comparing with that, we do have added the type checking result for function docs:

https://github.com/Myriad-Dreamin/tinymist/blob/5dd9badc112c1a8f864368261ef5b9692ac621bb/crates/tinymist-query/src/analysis/global.rs#L457-L462

Eric-Song-Nop:
Yes, thus I am calling `literal_type_of_node` on hover which I believe will do the type check. I am not changing the global, as I am not sure if that will break the type checker as there the return_ty should be the varType

Eric-Song-Nop:
Btw, I hesitate if this kind of type hint is possible for Typst as a dynamic lang...

crates/tinymist-query/src/hover.rs
```diff
@@ -178,7 +178,21 @@ fn def_tooltip(
 
                         match &sym_docs {
                             Some(DefDocs::Variable(docs)) => {
-                                push_result_ty(def.name(), docs.return_ty.as_ref(), &mut type_doc);
+                                if let Some((_, long, _)) = docs.return_ty.as_ref() {
+                                    if long.as_str() == format!("@{}", def.name()) {
```
Myriad-Dreamin, line 182: well, you compared type by formatted result😄. The formatted type must not be used for equality checking, which is limitation to solve.

Myriad-Dreamin:
When I looked into bad cases, I found I forgot to simplify types before formatting, I also put some docs to tell the reason.

Myriad-Dreamin:
> Btw, I hesitate if this kind of type hint is possible for Typst as a dynamic lang...

The type hints and the checker will be ultimately better than typescript one, at least not worse.

Eric-Song-Nop:
> When I looked into bad cases, I found I forgot to simplify types before formatting, I also put some docs to tell the reason.

What exactly is simplify? I don't see this word in type checking commonly.

Eric-Song-Nop:
Oh I see it's Canonicalize

Eric-Song-Nop:
Do you mind add some doc for simplify? especially what is the `principle` parameter? when should it be true?

Myriad-Dreamin:
> Oh I see it's canonicalize

People might use simplify or normalize to indicate the process. It isn't exactly the canonicalization or normalization. You know a type can be represented into different form. On the strictness, canonicalization > normalization > simplification. canonicalization requires the resulting form must be unique per type. normalization doesn't require uniqueness but must be fully evaluated. simplification just doesn't contain much formal property, but just make things simpler. In tinymist, it just further evaluates types depending the context, e.g. type aliases and gives more opportunities to make a shorter form. For example, `@a | @b` where `a` and `b` are both `int` in the context, it will be evaluated as `int | int` and then shorten as `int`. I didn't use `normalize` here as I'm not sure whether the related property is met here.

Myriad-Dreamin (approved):

PR #962, opened 2024-12-07
fix: import item complete

Eric-Song-Nop:
fix: `import "lib.typ": |` cannot complete

problem from #952

### DIFF
--- a/crates/tinymist-query/src/fixtures/completion/import_item.typ
+++ b/crates/tinymist-query/src/fixtures/completion/import_item.typ
@@ -3,4 +3,8 @@
 
 -----
 /// contains: base,aa
-#import "base.typ": /* range -1..1 */
+#import "base.typ":      
+/* range -3..2 */
+
+
+some text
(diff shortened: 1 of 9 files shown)


Myriad-Dreamin:
Still cannot complete directly after the colon:

```typ
#import "../src/basic/mod.typ":| 
```

image

Eric-Song-Nop:
seems there has to be a space after the colon

Myriad-Dreamin:
I fixed it by adding another matcher on colon marker, but I then find it does not work after comma as well.

```typ
#import "../src/basic/mod.typ": code,| 
```

Myriad-Dreamin:
it is probably nice enough. The only missing thing is to add more testing about the colon and comma completion.

Eric-Song-Nop:
> it is probably nice enough. The only missing thing is to add more testing about the colon and comma completion.

I made some

Myriad-Dreamin (approved):

PR #966, opened 2024-12-08
fix: resolve #986 avoid wrong offset on Neovim for some postfix snippets

### LINKED ISSUE #986: Some postfix snippets are applied in wrong order in Neovim
**Describe the bug**
`#[]. text fille` becomes:
`#tex, [])t(fill:`

Some snippets are inseerted or applied in wrong order

**Package/Software version:**

tinymist extension version:
```plain
tinymist 
Build Timestamp:     2024-12-11T14:56:58.376451561Z
Build Git Describe:  v0.12.10-21-g5747dd6-dirty
Commit SHA:          5747dd6ba623e5413fbd44f4ea64089e8a4eff00
Commit Date:         None
Commit Branch:       None
Cargo Target Triple: x86_64-unknown-linux-gnu
Typst Version:       0.12.0
Typst Source:        git+https://github.com/Myriad-Dreamin/typst.git?tag=tinymist-v0.12.0#58426a90a7ef721738a01be09793d33e55eb75a9
```

Neovim version:
```plain
NVIM v0.11.0-dev-1325+g30726c778c
Build type: RelWithDebInfo
LuaJIT 2.1.1732813678
Run "nvim -V1 -v" for more info
```

Eric-Song-Nop:
resolve #986

image1
image2

According to lsp spec, additional text edit is not supposed to be used here.

> /**
> 	 * An optional array of additional text edits that are applied when
> 	 * selecting this completion. Edits must not overlap (including the same
> 	 * insert position) with the main edit nor with themselves.
> 	 *
> 	 * Additional text edits should be used to change text unrelated to the
> 	 * current cursor position (for example adding an import statement at the
> 	 * top of the file if the completion item will insert an unqualified type).
> 	 */


Eric-Song-Nop:
Other uses of `additional_text_edits` should also be replaced gradually

crates/tinymist-query/src/upstream/complete/ext.rs
```diff
@@ -257,17 +260,25 @@ impl CompletionContext<'_> {
...
+                kind: Some(CompletionItemKind::SNIPPET),
+                label: snippet.label.clone().into(),
+                detail: Some(snippet.description.clone().into()),
+                insert_text_format: Some(InsertTextFormat::SNIPPET),
+                ..Default::default()
+            };
             if let Some(node_before_before_cursor) = &node_before_before_cursor {
                 let node_content = node.get().clone().into_text();
                 let before = TextEdit {
                     range: self.ctx.to_lsp_range(rng.start..self.from, &src),
```
Myriad-Dreamin, line 273: The range of main edit can only start at the cursor, but `rng.start` is before the cursor.

Myriad-Dreamin: Probably not, the lsp spec says, "it must be a [single line] and they must contain the position at which completion has been requested," rather than "start with the position".

Eric-Song-Nop: Yes, that's why it is ended in self.from.

Eric-Song-Nop: Maybe I should use the insertandreplace capability of lsp

Eric-Song-Nop:
I am not sure if this is a common problem, so I am turning this pr in to a specific one for #986

crates/tinymist-query/src/upstream/complete/ext.rs
Myriad-Dreamin: As the lsp spec said above, `rng.start..self.from` may span across multiple lines but the spec requires it MUST be in a single line so I don't think this is a valid fix. Btw, I checked it in my local neovim, and at least the case in #986 can work as expected. It might be worth to checking whether some neovim completion plugins don't work while others do work. I use the default LazyVim plus the `tinymist.lua` from [the commit](https://github.com/Myriad-Dreamin/tinymist/pull/993).

Myriad-Dreamin (approved):

PR #994, opened 2024-12-12
feat: complete field accesses by type information

### LINKED ISSUE #985: Complete field accesses by type information
**Motivation**

Tinymist should be able to get completion information from the type check information, instead of only the value of a variable.

**Description**

Type inference in Tinymist might provide more info than the value info from Typst, thus we could benefit from it for autocomplete to provide for information for completion, especially for field access completion

**Examples/Questions**
```typst
/// Function
///
/// - dic (dictionary, float): The argument dic
/// - val (string): 2nd arg
/// -> float
#let func(
  dic: (:),
  val: "",
) = return dic.|
```

For now, I believe no completion info can is provided from tinymist, while it could get the type of `dic` and provide field completion.

```typst
/// Function
///
/// - dic (dictionary, float): The argument dic
/// - val (string): 2nd arg
/// -> float
#let func(
  dic: (:),
  val: "",
) = return dic.|

#func(dic: (1, 2, 3))
```
While in this case, now it will complete as if `dic` is of value `(1, 2, 3)`, while the proper way might still be complete as (dictionary, float).

Eric-Song-Nop:
resolve #985 

This pr is limited to field access completion.

### DIFF
--- a/crates/tinymist-query/src/analysis/global.rs
+++ b/crates/tinymist-query/src/analysis/global.rs
@@ -813,7 +813,7 @@ impl SharedContext {
         self.type_check(source).type_of_span(s)
     }
 
-    pub(crate) fn literal_type_of_node(self: &Arc<Self>, k: LinkedNode) -> Option<Ty> {
+    pub(crate) fn literal_type_of_node(self: &Arc<Self>, k: &LinkedNode) -> Option<Ty> {
         let id = k.span().id()?;
         let source = self.source_by_id(id).ok()?;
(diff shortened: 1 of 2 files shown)


Myriad-Dreamin:
I'm not sure what's your plan but we have identified the target and field of field accesses for type-based completion just now.

https://github.com/Myriad-Dreamin/tinymist/blob/980571ad612e1581d42be3ae80dfd514ae2c7a8b/crates/tinymist-query/src/upstream/complete/ext.rs#L1400-L1405

Issue #1015, opened 2024-12-17
Auto Completion Tidy style doc comment


Eric-Song-Nop:
**Motivation**

Doc comment is handy in Typst, while repeaty to repeat the parameters. We can have tinymist to generate a doc comment snippet for use when insert `///`

**Description**

Similar to [VSCode auto JSDoc](https://code.visualstudio.com/updates/v1_10#_auto-jsdoc-comments).

**Examples/Questions**

when insert `///` before a function, we can generate a snippet for doc comment:

```typst
///
  ^
  |
#let sinc(x) = if x == 0 { 1 } else { calc.sin(x) / x }

// generate:

/// {$3 Desc}
///
/// - x ({$1}): $2
/// -> {$3}
#let sinc(x) = if x == 0 { 1 } else { calc.sin(x) / x }
```

Eric-Song-Nop:
（if this feature is wanted, I can do it

Eric-Song-Nop:
For me, Tinymist is the right place for this feature. Other doc string generators might struggle to implement this feature because of lack of high quality typst, for example treesitter parser for Typst behaves poorly for identifying function definitions, thus hard to implement [Neogen](https://github.com/danymat/neogen?tab=readme-ov-file#adding-languages) for Typst.

Myriad-Dreamin:
Hey, you found a good feature. First we could split two parts to discuss with less arguments:
- Q1: How does tinymist expose the function to users?
- Q2: How does tinymist-query provide a function to help that?

For Q1, instead of guessing people's mind, what about using code actions? It should provide a code action to toggle comment update on head of the let binding items (e.g. `#let f` of `#let f(x) = x`). Compared with always generate docs automatically, providing a such code action would be less controversial. 

For Q2, tinymist-query should provide a comment updater that not only generating docs from scratch but also updating existing docs. As a good start, it should be good enough to provide a simple generator.

PR #1029, opened 2024-12-19
feat: comment complete for tidy style doc comment

Eric-Song-Nop:
for #1015

### DIFF
--- a/crates/tinymist-query/src/upstream/complete.rs
+++ b/crates/tinymist-query/src/upstream/complete.rs
@@ -8,7 +8,7 @@ use lsp_types::TextEdit;
 use serde::{Deserialize, Serialize};
 use typst::foundations::{fields_on, format_str, repr, Repr, StyleChain, Styles, Value};
 use typst::model::Document;
-use typst::syntax::ast::AstNode;
+use typst::syntax::ast::{AstNode, Param};
 use typst::syntax::{ast, is_id_continue, is_id_start, is_ident, LinkedNode, Source, SyntaxKind};
 use typst::text::RawElem;
 use typst::World;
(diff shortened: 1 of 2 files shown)


Eric-Song-Nop:
For now this pull request just complete a doc string as comment complete, it is not considering the known type of the parameters, while from my perspective works better when this feature is implemented as a docstring updater or a code action.

Myriad-Dreamin:
I start to experience it and have found several points. We might start to complete a "Document function" only if the comment is a "doc comment" and only contains nothing.
- if user only enters two slashes, they might not purpose to make a documentation.
- we seems to be triggering the completion in the comment everywhere, e.g. when I put a dot in the comment.
   image
- we can also generates documentation for variables.

Eric-Song-Nop:
> if user only enters two slashes, they might not purpose to make a documentation.
> we seems to be triggering the completion in the comment everywhere, e.g. when I put a dot in the comment.

I have stricted the condition. Tbh, why doc comment does not utilize block comment but line comment, it's really annoying.

> we can also generates documentation for variables.
If I know correctly, doc a variable has no specific format right? Do we really need to complete anything for it?

Myriad-Dreamin:
> Tbh, why doc comment does not utilize block comment but line comment, it's really annoying.

Caused that we borrow the format from rust.

> If I know correctly, doc a variable has no specific format right? Do we really need to complete anything for it?

You might know it wrongly. You can document any let binding items.

```typ
/// A swap operation. 
///
/// - a (B): It was b.
/// - b (A): It was a.
/// -> (B, A)
#let (a, b) = (b, a);


/// A simpler example. 
///
/// -> content
#let embedded-file = context metadata(query(<output>));
```

Myriad-Dreamin (approved):

Eric-Song-Nop:
> You might know it wrongly. You can document any let binding items.
Oh I didn't know this

PR #1043, opened 2024-12-21
feat: fold continue line comments

Eric-Song-Nop:
resolve: #1044

Should be ready


Myriad-Dreamin:
I assigned the kind of these folding range to `Fol`, to allow us folding/unfolding all comments by one shortcut.

Myriad-Dreamin (dismissed):

Myriad-Dreamin:
We start to maintain some magic mutable state when calculating folding range. This is a signal that the code starts to become unmaintainable. 

```rs
let mut last_line_comment_end_range: Option<FoldingRange> = None;
```

This won't block us to merge this PR because there is an only state related to the conversion loop, but we may need some refactor to identify them uniformally in `LexicalHierarchyWorker`, where we can hold some analysis state in the `Worker` legally.

hint: Conventionally, every struct ends with `Worker` is an independent analyzer.

Eric-Song-Nop:
I am also thinking of getting rid of it, how I really cannot fully understand how the `lexicalhierechy` works clearly, thus I am doing all works in folding_range.

Myriad-Dreamin:
I have tried to refactor it a bit: https://github.com/Eric-Song-Nop/tinymist/compare/fold_line_comments...Myriad-Dreamin:fold_line_comments

I mainly just avoid pushing `Space`s to the stack and reuse some stuff from the existing comment matcher.

Eric-Song-Nop:
I've tested and it looks good to me.

Myriad-Dreamin (approved):

Eric-Song-Nop:
Might need to be merged quickly before other codes changed way too much.

PR #1047, opened 2024-12-21
chore: bump typstyle to v0.12.13

Enter-tainer:


crates/tinymist/src/task/format.rs
```diff
@@ -39,8 +40,15 @@ impl FormatTask {
...
-                    let res = typstyle_core::Typstyle::new_with_src(src.clone(), cw).pretty_print();
-                    Ok(calc_diff(src, res, c.position_encoding))
+                    let res = typstyle_core::Typstyle::new_with_src(
+                        src.clone(),
+                        PrinterConfig::new_with_width(cw),
+                    )
+                    .pretty_print();
+                    match res {
+                        Ok(res) => Ok(calc_diff(src, res, c.position_encoding)),
+                        Err(_) => Ok(None),
```
Enter-tainer, line 50: currently the only err is syntax error. pls check if this makes sense to you

Myriad-Dreamin: It looks ok to me, but I converted the `match` to a single `.ok()` to ignore the error.

Myriad-Dreamin (approved):

PR #1086, opened 2024-12-28
docs: state `shiroa` as a requirement

SylvanFranklin:
Building the docs via `yarn docs` wraps `shiroa`, I could not build them without first installing `shiroa` which should be stated clearly in the development guide.


Myriad-Dreamin:
I edit the tip a bit, is it okay?

Myriad-Dreamin (approved):

SylvanFranklin:
Yeah looks good

PR #1100, opened 2025-01-02
feat: support convert to typst table from xlsx file

hongjr03:
Hi! This pr inherits the implementation of [excel-to-typst repo](https://github.com/hongjr03/excel-to-typst) into tinymist. Due to the limitations of the vscode API, I was only able to read the user's clipboard in HTML format via a shell call from the platform.

I have tested this on my MacBook (by developing a vscode extension), the Linux platform is not yet tested. On Windows platforms there may be problems with character encoding (UTF8 -> GBK -> UTF8) due to calls to powershell, not yet solved.

### DIFF
--- a/editors/vscode/src/features/drag-and-drop.ts
+++ b/editors/vscode/src/features/drag-and-drop.ts
@@ -86,7 +86,7 @@ export class TextProvider implements vscode.DocumentDropEditProvider {
         codeSnippet = `grayscale-image(read(${strPath}))`;
         break;
       case ResourceKind.Xlsx:
-        additionalPkgs.push(["@preview/rexllent", "0.2.1", "xlsx-parser"]);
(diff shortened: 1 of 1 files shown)


Myriad-Dreamin:
For clipboard access, you could check `vscode.env.clipboard`, which is compatible with the [Electron clipboard](https://www.electronjs.org/docs/latest/api/clipboard).

hongjr03:
I'm wondering whether we can provide a paste and auto-completion method based on the power of [rexllent](https://github.com/hongjr03/typst-ReXLlenT).

Something like:
1. copy an excel file
2. paste in vscode window
3. automatically copy file to current working dir then auto complete package parsing code

hongjr03:
just found a bug and have to wait until 0.2.1 be merged :(

hongjr03:
0.2.1 have been merged :)

PR #1110, opened 2025-01-04
feat: Rescaling with Ctrl+=/- in browser (in addition to ctrl+wheel)

tmistele:
In normal browsers, one can zoom in and out by `ctrl`+`wheel` and also by `ctrl`+`=` or `ctrl`+`-` . tinymist so far only supports the former. This pr adds support for the latter.

### DIFF
--- a/tools/typst-dom/src/typst-doc.mts
+++ b/tools/typst-dom/src/typst-doc.mts
@@ -148,7 +148,7 @@ export class TypstDocumentContext<O = any> {
       this.hookedElem.parentElement?.classList.add("hide-scrollbar-y");
     }
 
-    this.installCtrlWheelHandler();
+    this.installRescaleHandler();
   }
 
   reset() {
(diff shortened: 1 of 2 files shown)


SylvanFranklin:
for the macOS is the control represented with command?

tmistele:
> for the macOS is the control represented with command?

If I'm reading the `KeyboardEvent` documentation <sup>[1] [2]</sup> correctly, no. The `KeyboardEvent.ctrlKey` property we're using is the "control" key. The mac "command" key seems to be `KeyboardEvent.metaKey`.

[1]: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/metaKey
[2]: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/ctrlKey

Enter-tainer:
The code overall looks good to me. Just want to confirm it doesn't change original behavior in both browser and vscode webview.

@Myriad-Dreamin I hope I can review and test this PR by hand but I'm too occupied by work recently.

SylvanFranklin:
> If I'm reading the `KeyboardEvent` documentation [1](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/

Does it make sense / is it logistically possible to make it command? It is more consistent with the behavior of web browsers on mac. Currently `command` `+` increases the zoom level without zooming the page, but maybe it is possible to tap into and override? If not it's probably not a huge problem.

tmistele:
> Does it make sense / is it logistically possible to make it command? It is more consistent with the behavior of web browsers on mac. Currently `command` `+` increases the zoom level without zooming the page, but maybe it is possible to tap into and override? If not it's probably not a huge problem.

Yes, it should be possible to use `metaKey` for Macs and `ctrlKey` otherwise. I can implement that.

What about the existing `Ctrl`+`wheel` behavior? Should that also be `command` instead of `ctrl` on macs?

SylvanFranklin:
> What about the existing Ctrl+wheel behavior? Should that also be command instead of ctrl on macs?

I'm not sure about that one since it's not as established as a keybind, I guess it would make sense to use command just for the sake of consistency.

tmistele:
I just checked the documentation for firefox: The default shortcuts for zooming on Mac are `Ctrl`+`Wheel` and `Command` + `=` / `Command` + `-`. So I guess that's what we should do.

https://support.mozilla.org/en-US/kb/mouse-shortcuts-perform-common-tasks
https://support.mozilla.org/en-US/kb/keyboard-shortcuts-perform-firefox-tasks-quickly

tmistele:
Ok, I've changed ctrl to command for Mac. I don't have a Mac to test this with, though. Can you test it @SylvanFranklin ?

SylvanFranklin (approved):

Myriad-Dreamin:
I should schedule a time to run it by myself, and merge it If no obvious issue is found.

SylvanFranklin:
@tmistele Did not work for me when I built with cargo. I tried it in both firefox and brave but did not get it to work.

tmistele:
> @tmistele Did not work for me when I built with cargo. I tried it in both firefox and brave but did not get it to work.

Did you run `yarn build:preview` and then uncomment this line @SylvanFranklin  ?

https://github.com/Myriad-Dreamin/tinymist/blob/8e918cb1322c67160a516c7478d6ecd7a444f014/Cargo.toml#L217

I think otherwise you will be using the version of `tinymist-assets`  that's on crates.io

SylvanFranklin:
Yes, thanks for the clarification, should have read the build docs more closely. This now works for me on both firefox and chromium, well done.

Myriad-Dreamin:
Thanks you both! I'll merge it without running it because it looks good to me and @SylvanFranklin has confirmed it works smoothly. I'll finally check it by myself when we come to a new release.

A small improvement is that we could list the new key bindings in the help panel to let people check them, in case people bind their browser's zoom shortcuts to different ones.

https://github.com/Myriad-Dreamin/tinymist/blob/ab0903934de3b04644c69af8c8dd023b53b345e6/tools/typst-preview-frontend/index.html#L266

Myriad-Dreamin (approved):

PR #1156, opened 2025-01-12
feat: remove the typst actor

Myriad-Dreamin:
We now benefit from having a clonable `typst::World`, i.e. the compile snapshot. Replace the typst actor with a stateless `EditorServer`. The span resolution is handled by the compile snapshot. With this change, the typst preview server can solve the spans synchronously and send the memory changes to the language server directly.


crates/typst-preview/src/actor/render.rs
```diff
@@ -63,43 +68,40 @@ impl RenderActor {
...
         let res = msg.is_full_render();
 
         match msg {
+            RenderActorRequest::DocToSrcJumpResolve(span_range) => {
+                log::debug!("RenderActor: processing doc2src: {span_range:?}");
+
+                self.doc_to_src_jump_resolve(span_range);
+            }
+            RenderActorRequest::ChangeCursorPosition(req) => {
+                log::debug!("RenderActor: processing src2doc: {:?}", req);
```
Myriad-Dreamin, line 82: There was some copy typo.

Issue #625, opened 2024-10-03
Websocket connection for preview on code-server


Checker8763:
**Describe the bug**

When using the extension on code-server the preview tries to connect to localhost instead of the proxy url.

**Description**

It would be nice to set the websocket path for the lsp to connect to because automatically getting it seems complex or impossible becasue it is configurable in code-server.

**Examples/Questions**

- Becasue the proxy url can be configured to be a subpath or a subdomain it would be nice to change to either method with a switch (Radio button) and configure it
  1. Localhost
  2. Subdomain
  3. Subpath
- Raising questions to highlight the uncertain design for the *feature*.

**Package/Software version:**

code-server version(Help -> About):
docker image: codercom/code-server:4.93.1-ubuntu
```plain
code-server: v4.93.1
Code: 1.93.1
Commit: 3542bd157b0ccc8bf732acdb03d4d69d85f48ea9
Date: 2024-09-19T10:15:30.023Z
Browser: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/124.0.0.0 Safari/537.36
```

tinymist extension version: `v0.11.22`. Get it by `tinymist --version` in terminal.
as bundeled with the extension of this version
```plain
tinymist 
Build Timestamp:     2024-09-27T23:50:57.941849314Z
Build Git Describe:  v0.11.22
Commit SHA:          f3ccb4a18618fcf51d2ada8217629c58783e338c
Commit Date:         None
Commit Branch:       None
Cargo Target Triple: x86_64-unknown-linux-gnu
Typst Version:       0.11.1
```

**Logs:**

tinymist server log(Output Panel -> tinymist):
```plain
[Trace - 9:20:36 AM] Sending request 'initialize - (0)'.
[2024-10-03T09:20:36Z INFO  tinymist] starting LSP server: LspArgs {
        mirror: MirrorArgs {
            mirror: "",
            replay: "",
        },
        font: CompileFontArgs {
            font_paths: [],
            ignore_system_fonts: false,
        },
    }
[2024-10-03T09:20:36Z INFO  sync_lsp] handling initialize - (0) at Instant { tv_sec: 5430869, tv_nsec: 252669164 }
[...]

PR #1157, opened 2025-01-12
fix: Prevent malicious websites from connecting to http / websocket server

tmistele:
Simpler alternative to #1093 that does not break any downstream packages.

However, this does _not_ provide protection on multi-user systems where users don't trust each other. It only prevents websites other than `127.0.0.1` from connecting to our servers. So it's a good first step but doesn't cover all scenarios.

One question:

I noticed there's a second `make_http_server` function that explicitly sets `Access-Control-Allow-Origin: *`. What is that server used for? Obviously, the `*` CORS header looks dangerous.

https://github.com/Myriad-Dreamin/tinymist/blob/8e918cb1322c67160a516c7478d6ecd7a444f014/crates/tinymist/src/task/user_action.rs#L227-L264

### DIFF
--- a/crates/tinymist/src/task/user_action.rs
+++ b/crates/tinymist/src/task/user_action.rs
@@ -245,14 +245,25 @@ pub async fn make_http_server(
             let timings = timings.clone();
             let _ = alive_tx.send(());
             async move {
+                // Make sure VSCode can connect to this http server but no malicious website a user
+                // might open in a browser. We recognize VSCode by an `Origin` header that starts
+                // with `vscode-webview://`. Malicious websites can (hopefully) not trick browsers
+                // into sending an `Origin` header that starts with `vscode-webview://`
+                //
+                // See comment in `make_http_server` in `crates/tinymist/src/tool/preview.rs` for more
+                // details. In particular, note that this does _not_ protect against malicious users
+                // that share the same computer as us.
+                let Some(allowed_origin) = req
+                    .headers()
+                    .get("Origin")
+                    .filter(|h| h.as_bytes().starts_with(b"vscode-webview://"))
+                else {
+                    anyhow::bail!("Origin must start with vscode-webview://");
+                };
+
+                let b = hyper::Response::builder()
+                    .header(hyper::header::ACCESS_CONTROL_ALLOW_ORIGIN, allowed_origin);
                 if req.uri().path() == "/" {
-                    let b = hyper::Response::builder()
-                        .header(hyper::header::ACCESS_CONTROL_ALLOW_ORIGIN, "*")
-                        .header(hyper::header::ACCESS_CONTROL_ALLOW_METHODS, "GET, HEAD")
-                        .header(
-                            hyper::header::ACCESS_CONTROL_ALLOW_HEADERS,
-                            "Origin, X-Requested-With, Content-Type, Accept",
-                        );
                     let res = if req.method() == hyper::Method::HEAD {
                         b.body(Full::<Bytes>::default()).unwrap()
                     } else {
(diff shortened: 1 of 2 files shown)


tmistele:
I just realized this probably won't work like this for vscode because vscode sends something like `vscode-webview://<random-stuff>` as `Origin` header so we can't just check `Origin == 127.0.0.1:port`. But it might be enough to check `Origin` == 127.0.0.1:port || `Origin` starts with "vscode-webview://`". I'll have to think about it (don't have more time to work on it right now, maybe later today).

Myriad-Dreamin:
I added it because VS Code rejects fetch data from localhost. It should be good with the only place to use and it should be better to put down all allowed origins one by one instead of a simple wildcard. We should change the code in a separated PR.

https://github.com/Myriad-Dreamin/tinymist/blob/8e918cb1322c67160a516c7478d6ecd7a444f014/tools/editor-tools/src/features/tracing.ts#L127

Myriad-Dreamin:
The new PR is much simpler, although it only authenticates the client by the origin. 

Just a small question remaining in my view. Will the check of the origin break people who use http proxy of code servers like gitpod and codeserver? These servers may only forward the http request without changing the origin.

- @tani, the author of the [gitpod.ts](https://github.com/Myriad-Dreamin/tinymist/blob/6813a1081e37e53e64cc1663da8daa033c9e7bc0/editors/vscode/src/gitpod.ts#L15-L32)
- @ because of https://github.com/Myriad-Dreamin/tinymist/issues/625

https://github.com/Myriad-Dreamin/tinymist/blob/6813a1081e37e53e64cc1663da8daa033c9e7bc0/editors/vscode/src/gitpod.ts#L14-L32

tmistele:
> I added it because VS Code rejects fetch data from localhost. It should be good with the only place to use and it should be better to put down all allowed origins one by one instead of a simple wildcard. We should change the code in a separated PR.

I've added a commit that I think addresses this, but please have a look @Myriad-Dreamin . This allows only VSCode to connect to the trace/profiling http server for now.

Myriad-Dreamin:
> The new PR is much simpler, although it only authenticates the client by the origin.
> 
> Just a small question remaining in my view. Will the check of the origin break people who use http proxy of code servers like gitpod and codeserver? These servers may only forward the http request without changing the origin.
> 
> * @tani, the author of the [gitpod.ts](https://github.com/Myriad-Dreamin/tinymist/blob/6813a1081e37e53e64cc1663da8daa033c9e7bc0/editors/vscode/src/gitpod.ts#L15-L32)
> * @ because of [Websocket connection for preview on code-server #625](https://github.com/Myriad-Dreamin/tinymist/issues/625)
> 
> https://github.com/Myriad-Dreamin/tinymist/blob/6813a1081e37e53e64cc1663da8daa033c9e7bc0/editors/vscode/src/gitpod.ts#L14-L32

This is the final blocking part. I guess we should allow the requests from `url.hostname` but I don't know how to test it:

```ts
 url.hostname = 
   url.port.toString() + 
   "-" + 
   process.env.GITPOD_WORKSPACE_ID + 
   "." + 
   process.env.GITPOD_WORKSPACE_CLUSTER_HOST; 
```

@tani @tmistele

tmistele:
> This is the final blocking part. I guess we should allow the requests from `url.hostname` but I don't know how to test it:
> 
> ```ts
>  url.hostname = 
>    url.port.toString() + 
>    "-" + 
>    process.env.GITPOD_WORKSPACE_ID + 
>    "." + 
>    process.env.GITPOD_WORKSPACE_CLUSTER_HOST; 
> ``

Could be. But I don't know how gitpod works, so I'm not sure.

Is this essentially a version of vscode that runs on a gitpod server and renders its output to a browser? If so, maybe it sends a `vscode-webview://...` URL as `Origin` header when connecting to the websocket, just like the "normal" vscode does? So the current version of this PR might already work with gitpod?

Myriad-Dreamin (approved):

PR #1158, opened 2025-01-12
fix(vscode): make label view work when there's exactly one label

tmistele:
The labels tool view in vscode didn't work for documents with exactly one label. The error "items is not an iterable" was shown. This PR fixes that.

Before:

image

After:

image


Myriad-Dreamin:
Thank you. I tested it on the `editors/vscode/e2e-workspaces/issue-1158/label-test.typ`, and the patch does fix the bug.

Myriad-Dreamin (approved):

PR #1259, opened 2025-02-03
build: downgrade `tempfile` to 3.15.0

Myriad-Dreamin:


CGMossa:
Is there something wrong with tempfile?

PR #1282, opened 2025-02-13
feat: better contextual render

hongjr03:
- Better contexual render

    I introduced an `EcoString` to store the codes of nodes that result in `Value::None` (excluding `Hash | Ident | Bool | Int | Float | Numeric | Str | Array | Dict`) that appear before a `Contextual` node. Then prepend the buffer to the `Contextual` node to render in right environment.
    
    Add assets_path and assets_src_path to args and TypliteFeat, support render contexual code to external svg file.

Variable names were chosen quickly and are not ideal apparently, sorry for that :(

### DIFF
--- a/crates/typlite/src/lib.rs
+++ b/crates/typlite/src/lib.rs
@@ -421,15 +421,9 @@ impl TypliteWorker {
             ""
         };
 
-        let write_error = |content: &mut EcoString, align: &str, err: &str, inline: bool| {
-            if !inline {
-                let _ = write!(content, r#"<p align="{align}">"#);
-            }
+        let write_error = |content: &mut EcoString, err: &str| {
             let err = err.replace("`", r#"\`"#);
             let _ = write!(content, "```\nRender Error\n{err}\n```");
-            if !inline {
-                content.push_str("</p>");
-            }
         };
 
         let write_image = |content: &mut EcoString,
@@ -480,7 +474,7 @@ impl TypliteWorker {
                 let data = match render(theme) {
                     Ok(data) => data,
                     Err(err) if self.feat.soft_error => {
-                        write_error(&mut content, align, &err.to_string(), inline);
+                        write_error(&mut content, &err.to_string());
                         return Ok(Value::Content(content));
                     }
                     Err(err) => return Err(err),
@@ -516,7 +510,7 @@ impl TypliteWorker {
                 let dark = match render(ColorTheme::Dark) {
                     Ok(d) => d,
                     Err(err) if self.feat.soft_error => {
-                        write_error(&mut content, align, &err.to_string(), inline);
+                        write_error(&mut content, &err.to_string());
                         return Ok(Value::Content(content));
                     }
                     Err(err) => return Err(err),
@@ -524,7 +518,7 @@ impl TypliteWorker {
                 let light = match render(ColorTheme::Light) {
                     Ok(l) => l,
                     Err(err) if self.feat.soft_error => {
-                        write_error(&mut content, align, &err.to_string(), inline);
+                        write_error(&mut content, &err.to_string());
                         return Ok(Value::Content(content));
                     }
                     Err(err) => return Err(err),
--- a/crates/typlite/src/lib.rs
+++ b/crates/typlite/src/lib.rs
@@ -310,7 +310,12 @@ impl TypliteWorker {
             LineComment => Ok(Value::None),
             BlockComment => Ok(Value::None),
         };
-        if res.clone()? == Value::None && !matches!(node.kind(), Hash | Ident) {
+        if res.clone()? == Value::None
+            && !matches!(
+                node.kind(),
+                Hash | Ident | Bool | Int | Float | Numeric | Str | Array | Dict
+            )
+        {
             self.pref += node.clone().into_text();
             self.pref += "\n";
         }
--- a/crates/typlite/src/lib.rs
+++ b/crates/typlite/src/lib.rs
@@ -114,6 +114,7 @@ impl Typlite {
             current,
             feat: self.feat,
             list_depth: 0,
+            pref: EcoString::new(),
             scopes: self
                 .library
                 .as_ref()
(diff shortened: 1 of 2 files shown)


Myriad-Dreamin:
Could you move the fix [fix: add no-content-hint to default features](https://github.com/Myriad-Dreamin/tinymist/pull/1282/commits/eb6fbab3bc5bb91c3ecaea6e227dc25de3a5da95) to a separated PR?

crates/typlite/src/lib.rs
```diff
@@ -363,6 +399,17 @@ impl TypliteWorker {
         inline: bool,
     ) -> Result<Value> {
         let theme = self.feat.color_theme;
+
+        let code_file_name = if let Some(assets_src_path) = &self.feat.assets_src_path {
+            Some(format!(
```
Myriad-Dreamin, line 404: use `self.feat.assets_src_path.join(eco_format!(".."))`

crates/typlite/src/lib.rs
```diff
@@ -372,20 +419,58 @@ impl TypliteWorker {
...
-                            let _ = write!(content, r#"<p align="{align}">"#);
-                        }
-                        let _ = write!(
-                            content,
-                            r#"<img{inline_attrs} alt="typst-block" src="data:image/svg+xml;base64,{data}" {extra_attrs}/>"#,
-                        );
-                        if !inline {
-                            content.push_str("</p>");
+                        if let Some(assets_path) = &self.feat.assets_path {
+                            let file_name = format!(
```
Myriad-Dreamin, line 429: We have made 5 awkward indents, `match Some match Ok if let ..`. We might rewrite the code to reduce the indent level of code here

crates/typlite/src/lib.rs
```diff
@@ -402,15 +487,54 @@ impl TypliteWorker {
...
-                            let _ = write!(content, r#"<p align="{align}">"#);
-                        }
-                        let _ = write!(
-                            content,
-                            r#"<picture><source media="(prefers-color-scheme: dark)" srcset="data:image/svg+xml;base64,{dark}"><img{inline_attrs} alt="typst-block" src="data:image/svg+xml;base64,{light}" {extra_attrs}/></picture>"#,
-                        );
-                        if !inline {
-                            content.push_str("</p>");
+                        if let Some(assets_path) = &self.feat.assets_path {
+                            let dark_file_name = format!(
```
Myriad-Dreamin, line 491: use `path.join`

crates/typlite/src/lib.rs
Myriad-Dreamin, line 429: use `path.join`

crates/typlite/src/lib.rs
```diff
@@ -516,7 +640,21 @@ impl TypliteWorker {
         })?;
 
         let svg_payload = typst_svg::svg_merged(&document, Abs::zero());
-        Ok(base64::engine::general_purpose::STANDARD.encode(svg_payload))
+
+        if let Some(assets_path) = &self.feat.assets_path {
+            let file_name = format!(
```
Myriad-Dreamin, line 645: use `path.join`

crates/typlite/src/lib.rs
```diff
@@ -582,6 +720,15 @@ impl TypliteWorker {
     fn raw(node: &SyntaxNode) -> Result<Value> {
         let mut s = EcoString::new();
         let raw = node.cast::<ast::Raw>().unwrap();
+        if let Some(lang) = raw.lang() {
+            if &EcoString::from("typlite") == lang.get() {
```
Myriad-Dreamin, line 724: what is a `typlite` lang? This looks like a feature not belonging to "better contextual render".

crates/typlite/src/lib.rs
```diff
@@ -667,7 +814,7 @@ impl TypliteWorker {
             return self.to_raw_block(node, !equation.block());
         }
 
-        self.render(node, !equation.block())
+        self.render(&SyntaxNode::default(), node, !equation.block())
```
Myriad-Dreamin, line 817: `SyntaxNode::default()` looks strange and may be broken by typst silently, we may use `SyntaxNode::leaf(Text, "")` instead here.

crates/typlite/src/main.rs
```diff
@@ -20,6 +20,14 @@ pub struct CompileArgs {
     /// Path to output file
     #[clap(value_name = "OUTPUT")]
     pub output: Option<String>,
+
+    /// Configures the path of assets directory
+    #[clap(long, default_value = None, value_name = "ASSETS_PATH")]
+    pub assets_path: Option<String>,
+
+    /// Configures the path of assets' source code directory
```
Myriad-Dreamin, line 28: We may describe what's the usage of these two paths having similar names.

crates/typlite/src/main.rs
```diff
@@ -36,11 +44,43 @@ fn main() -> typlite::Result<()> {
         Some(output_path) => Some(PathBuf::from(output_path)),
         None => Some(Path::new(input).with_extension("md")),
     };
+    let assets_path = match args.assets_path {
+        Some(assets_path) => {
+            let path = PathBuf::from(assets_path);
+            if !path.exists() {
+                if let Err(e) = std::fs::create_dir_all(&path) {
+                    eprintln!("Failed to create assets directory: {}", e);
```
Myriad-Dreamin, line 52: Will rust print the error message for us?

Myriad-Dreamin, line 52: The error messages should be lowercased.

hongjr03: seems like it's useless... I'll remove it.

crates/typlite/src/lib.rs
hongjr03: should have move to a separated pr

hongjr03:
> Could you move the fix [fix: add no-content-hint to default features](https://github.com/Myriad-Dreamin/tinymist/pull/1282/commits/eb6fbab3bc5bb91c3ecaea6e227dc25de3a5da95) to a separated PR?

Sure! See https://github.com/Myriad-Dreamin/tinymist/pull/1297

PR #1295, opened 2025-02-16
fix: respect that the port of the `expected_origin` can be zero

Myriad-Dreamin:
CC @tmistele

### DIFF
--- a/crates/tinymist/src/tool/preview.rs
+++ b/crates/tinymist/src/tool/preview.rs
@@ -13,6 +13,7 @@ use hyper_tungstenite::{tungstenite::Message, HyperWebsocket, HyperWebsocketStre
 use hyper_util::rt::TokioIo;
 use hyper_util::server::graceful::GracefulShutdown;
 use lsp_types::notification::Notification;
+use lsp_types::Url;
 use parking_lot::Mutex;
 use reflexo_typst::debug_loc::SourceSpanOffset;
 use reflexo_typst::Bytes;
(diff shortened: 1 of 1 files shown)


crates/tinymist/src/tool/preview.rs
```diff
@@ -602,6 +606,24 @@ pub async fn make_http_server(
     }
 }
 
+fn is_valid_origin(h: &HeaderValue, expected_origin: &str, expected_port: u16) -> bool {
```
tmistele, line 609: Can you say a bit more about what the overall idea is @Myriad-Dreamin ? Is it to support both "localhost" and "127.0.0.1" (instead of just "127.0.0.1") as host? And https in addition to just http?

In any case, I think there's no reason to allow "ws" and "wss" as the protcol/scheme. Is there any scenario where allowing "ws" / "wss" is needed?

Myriad-Dreamin: It is a quick fix to allow origin from `{http,https,ws,wss}://{localhost,127.0.0.1}:the-port`. It is "quick" so we'll remove the hardcoded check to the "localhost" and "127.0.0.1" later. About allowing the `ws` and `wss`, I haven't checked whether it will occur in the origin header, I guess no, but I think it is not harmful to add it.

Myriad-Dreamin: Note it is not merged into the main branch. it is only on the [nightly/v0.12.21](https://github.com/Myriad-Dreamin/tinymist/tree/nightly/v0.12.21) branch.

PR #1296, opened 2025-02-16
feat: embed Markdown codes

hongjr03:
- Embed Markdown codes

    Raw codes with `typlite` language will not be treated as a code block but directly output into the Markdown result.

### DIFF
--- a/crates/typlite/src/lib.rs
+++ b/crates/typlite/src/lib.rs
@@ -582,6 +582,15 @@ impl TypliteWorker {
     fn raw(node: &SyntaxNode) -> Result<Value> {
         let mut s = EcoString::new();
         let raw = node.cast::<ast::Raw>().unwrap();
+        if let Some(lang) = raw.lang() {
+            if &EcoString::from("typlite") == lang.get() {
+                for line in raw.lines() {
+                    s.push_str(&Self::value(Self::str(line.to_untyped())?));
+                    s.push('\n');
+                }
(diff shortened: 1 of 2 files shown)


crates/tinymist-task/Cargo.toml
```diff
@@ -42,7 +42,7 @@ notify.workspace = true
 
 [features]
 
-default = ["pdf", "text"]
+default = ["pdf", "text", "no-content-hint"]
```
Myriad-Dreamin, line 45: revert the unrelated change.

crates/typlite/src/lib.rs
```diff
@@ -582,6 +582,15 @@ impl TypliteWorker {
     fn raw(node: &SyntaxNode) -> Result<Value> {
         let mut s = EcoString::new();
         let raw = node.cast::<ast::Raw>().unwrap();
+        if let Some(lang) = raw.lang() {
```
Myriad-Dreamin, line 585: It could be better to add some comment here:

> Raw codes with typlite language will not be treated as a code block but directly output into the Markdown result.

Also, I think it should be documented somewhere to tell the feature otherwise people won't find it is feasible to embed markdown code directly.

hongjr03: okay I'll mention it in readme

Myriad-Dreamin (approved):

PR #1321, opened 2025-02-19
fix: cleanup CORS checks

tmistele:
cc @Myriad-Dreamin I've tried to cleanup the code from https://github.com/Myriad-Dreamin/tinymist/pull/1295 . I don't think we need to allow `ws` and `wss`, so I removed that. Two questions:

- Do we need to support `https` here?
- In this version, I'm only allowing `localhost`/`127.0.0.1`. Was your idea to parse the `static_file_addr`, extract the host from that, and use that host to match against the `Origin` header?

### DIFF
--- a/crates/tinymist/src/tool/preview.rs
+++ b/crates/tinymist/src/tool/preview.rs
@@ -449,20 +449,28 @@ pub async fn make_http_server(
     use hyper::body::{Bytes, Incoming};
     type Server = hyper_util::server::conn::auto::Builder<hyper_util::rt::TokioExecutor>;
 
+    let expected_host = Url::parse(&format!("http://{static_file_addr}"))
+        .unwrap()
+        .host_str()
+        .unwrap()
+        .to_string();
     let listener = tokio::net::TcpListener::bind(&static_file_addr)
         .await
         .unwrap();
     let addr = listener.local_addr().unwrap();
     log::info!("preview server listening on http://{addr}");
+    // Don't take the port from `static_file_addr` (it may have a dummy port e.g. `127.0.0.1:0`)
     let expected_port = addr.port();
 
     let frontend_html = hyper::body::Bytes::from(frontend_html);
     let make_service = move || {
         let frontend_html = frontend_html.clone();
         let websocket_tx = websocket_tx.clone();
+        let expected_host = expected_host.clone();
         service_fn(move |mut req: hyper::Request<Incoming>| {
             let frontend_html = frontend_html.clone();
             let websocket_tx = websocket_tx.clone();
+            let expected_host = expected_host.clone();
             async move {
                 // When a user visits a website in a browser, that website can try to connect to
                 // our http / websocket server on `127.0.0.1` which may leak sensitive information.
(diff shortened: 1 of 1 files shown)


tmistele:
> > * Do we need to support `https` here?
> 
> What's the downside of including it? If not, we could allow both `http` and `https` here.

I think there is currently no way to connect to tinymist using https? So I don't see why we should bother putting it in here. For security-related things I think it's best practice to allow only what is really needed.

> > Was your idea to parse the `static_file_addr`, extract the host from that, and use that host to match against the `Origin` header?
> 
> We should further support it, because it is weird if somebody specifies `--static_file_addr preview.somebody.io:80` but they cannot access it by `http://preview.somebody.io:80` or `https://preview.somebody.io:443`. Since I doubt whether it is a legal use (although I got some feature request to access some preview server remotely), I also feel okay to not support it intermediately.

I have now implemented getting the host from `static_file_addr`. I think there's then no need to separately allow 127.0.0.1/localhost, so I removed them. The code now only allows the host from `static_file_addr`.

Myriad-Dreamin:
> I think there is currently no way to connect to tinymist using https? So I don't see why we should bother putting it in here. For security-related things I think it's best practice to allow only what is really needed.

No, as what I doubt in last PR, the gitpod users may not be able to use preview in future tinymist. We just didn't take care of them and merge the PR. Their referer HTTP header could be something like `https://port-workspace.gitpod.com`.

> > We should further support it, because it is weird if somebody specifies `--static_file_addr preview.somebody.io:80` but they cannot access it by `http://preview.somebody.io:80` or `https://preview.somebody.io:443`. Since I doubt whether it is a legal use (although I got some feature request to access some preview server remotely), I also feel okay to not support it intermediately.
> 
> I have now implemented getting the host from `static_file_addr`. I think there's then no need to separately allow 127.0.0.1/localhost, so I removed them. The code now only allows the host from `static_file_addr`.

I see you delete the check to `vscode-webview://`, is it a missing? Maybe we should write some tests to check our code case by case.

tmistele:
> No, as what I doubt in last PR, the gitpod users may not be able to use preview in future tinymist. We just didn't take care of them and merge the PR. Their referer HTTP header could be something like `https://port-workspace.gitpod.

I wonder how that works, given that tinymist itself does not support `https`. So for example, I think gitpod can't do `tinymist --host post-workspace.gitpod.com:443`. That would just start a normal `http` server on port 443. Maybe they do `tinymist --host 127.0.0.1:23625` and then set up a separate https proxy that passes all the traffic through to tinymist.

> I see you delete the check to `vscode-webview://`, is it a missing? Maybe we should write some tests to check our code case by case.

Ah yeah, that was by accident, I've just put it back. I haven't properly tested this PR yet (as you can tell). I agree some tests would be good. Is there any existing infrastructure to test the websocket/http servers already?

Btw, when I run `tinymist preview <some-existing-file.typ>` with this branch it doesn't actually seem to compile the file and doesn't render anything in the browser. The browser connects to the websocket but never receives any content to render. Is that expected? Is there anything I need to consider that has changed?

Myriad-Dreamin:
> Maybe they do `tinymist --host 127.0.0.1:23625` and then set up a separate https proxy that passes all the traffic through to tinymist.

Yeah, I cannot check all of the workflow but this should be the most common uses. No magic will be happen, they won't touch the code of tinymist extension and the content of http requests, so the CLI argument is usually the `127.0.0.1:23625`. @tani  requested to turn the host to that of the gitpod in #575.

> Ah yeah, that was by accident, I've just put it back. I haven't properly tested this PR yet (as you can tell). I agree some tests would be good. Is there any existing infrastructure to test the websocket/http servers already?

No, I'm also seeking one to test the preview server continously. To test it simply, we might only extract some functions and make some unit testing.

> Btw, when I run `tinymist preview <some-existing-file.typ>` with this branch it doesn't actually seem to compile the file and doesn't render anything in the browser. The browser connects to the websocket but never receives any content to render. Is that expected? Is there anything I need to consider that has changed?
> 
> Console output for `tinymist preview test.typ`

Sorry, it is exactly broken in CLI. I fixed it on main.

Issue #1319, opened 2025-02-18
Config option to set formatter tab width


AgentElement:
**Motivation**

`typstyle` [v0.12.15](https://github.com/Enter-tainer/typstyle/commit/265619098f9e9948d42e27425d9bfb8c29402c38) now has support for a `--tab-width` flag. I think that a user should be able to configure tinymist to set a value for this flag in their editor config.

**Description**

Add a new `formatterTabWidth` option to `init::CONFIG_ITEMS`. Pass the value directly to `typstyle` with the new `.with_tab_spaces()` method.

**Notes**
This feature will not be available for users of `typstfmt`. I do not think this is an issue, as `typstfmt` is archived and no longer accepts PRs.

***

Would this feature be appreciated? I intend to work on it, and it is a relatively small change.

PR #1324, opened 2025-02-19
build: upgrade typstyle to v0.12.15

Eric-Song-Nop:
complete #1319 also

### DIFF
--- a/crates/tinymist/src/init.rs
+++ b/crates/tinymist/src/init.rs
@@ -278,6 +278,7 @@ const CONFIG_ITEMS: &[&str] = &[
     "semanticTokens",
     "formatterMode",
     "formatterPrintWidth",
+    "formatterTabSpaces",
     "completion",
     "fontPaths",
     "systemFonts",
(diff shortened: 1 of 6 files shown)


crates/tinymist/src/init.rs
```diff
@@ -278,6 +278,7 @@ const CONFIG_ITEMS: &[&str] = &[
     "semanticTokens",
     "formatterMode",
     "formatterPrintWidth",
+    "formatterTabSpaces",
```
Myriad-Dreamin, line 281: I would prefer to use `formatterTabWidth`, which is aligned with prettier's `tabWidth`. The `formatterPrintWidth` is also borrowed from prettier's naming. https://prettier.io/docs/options#tab-width

I actually think it is more precise to use editorconfig's `indent_size`, so it is also a candidate name.

Eric-Song-Nop: Ident size sounds most presice to me, whille using same name with tystyle might cause less confusion from my opinion

Myriad-Dreamin: @QuarticCat @Enter-tainer the `indent_space` should be renamed to `indent_size` or `tab_width`. `indent_space` sounds like some randomly abbr of name: `indent_size_and_using_space`.

QuarticCat: While it's bike-shedding, I personally prefer keeping the same with typstyle. Just make sure we keep the description straight and clear.

FYI, in VSCode settings, `editor.tabSize` refers to the width of a TAB character, and `editor.indentSize` refers to the number of spaces for an indentation level.

Myriad-Dreamin (approved):

AgentElement:
Thanks, I think we can close #1319 now.

PR #1368, opened 2025-02-22
build: bump version to 0.13.0

Myriad-Dreamin:
Small note: v0.13+ use typst v0.13.0, while v0.12+ continue using typst v0.12.0. The v0.12+ will be maintainted until most people move to v0.13+ or it takes too much cost to backport patches.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -408,9 +408,9 @@ checksum = "37b2a672a2cb129a2e41c10b1224bb368f9f37a2b16b612598138befd7b37eb5"
 
 [[package]]
 name = "cc"
-version = "1.2.14"
+version = "1.2.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


MultisampledNight:
Hoi! Would it be possible to make the Typst fork at https://github.com/Myriad-Dreamin/typst/tree/tinymist/v0.13.0-rc1 build on the `v0.13.0` branch of Typst instead, so that only exactly one `typst-assets` is pulled?

Specifically, the `main` branch of Typst (as well as the mentioned fork) uses `typst-assets` as *git* dep, whereas tinymist uses the *registry* dep.
(Noticed this when preparing an update for nixpkgs, a vendoring helper script was confused and crashed due to 2 `typst-assets` on version `0.13.0` being present instead of just 1)

Myriad-Dreamin:
@MultisampledNight Hi, thank you for reviewing it. laurmaedje also pointed out that v0.13.0 is slightly different from the 0.13.0 on main. I compared them just now and find v0.13.0 contains some important compatibility commit. Great news is that we haven't bumped it yet.

PR #1382, opened 2025-02-23
fix: cleanup CORS checks (v2)

tmistele:
Supersedes https://github.com/Myriad-Dreamin/tinymist/pull/1321

Mostly just cleanup, with a few behavior changes:

- Allows only https for gitpod (not http)
- Disallows any paths in Origin header (see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin#syntax)
- Allow a mismatch in ports only for localhost/127.0.0.1
- Always allow requests without any `Origin` header (the idea being that browsers do always send an `Origin` when it's important, in the meantime that same behavior was also added in https://github.com/Myriad-Dreamin/tinymist/pull/1411)

### DIFF
--- a/crates/tinymist/src/tool/preview.rs
+++ b/crates/tinymist/src/tool/preview.rs
@@ -494,10 +494,11 @@ pub async fn make_http_server(
             let static_file_addr = static_file_addr.clone();
             async move {
                 // When a user visits a website in a browser, that website can try to connect to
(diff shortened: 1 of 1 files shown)


Myriad-Dreamin (approved):

Myriad-Dreamin:
I find there are many `unwrap` in the file. We may clean up them in another PR..

PR #1411, opened 2025-02-26
fix: pass origin checking anyway

Myriad-Dreamin:
This is only a test branch for typst-preview.nvim people. They use a client that doesn't set `Origin` correctly.


tmistele:
Agree that should be fine. Normal browsers always send an `Origin` header when it is important. The same behavior I had also implemented in https://github.com/Myriad-Dreamin/tinymist/pull/1382 , though I didn't mention it in the PR description, I should probably mention it there.

PR #1520, opened 2025-03-16
docs: fix Neovim name casing everywhere

Andrew15-5:
From the https://neovim.io/ it's clearly seen that the proper naming would be "Neovim" and not "NeoVim". This PR fixes the casing and also capitalizes the first letter wherever is necessary. I would've automated it, but there are a lot of places where it's used as a part of a path etc., so I went through all files manually.


Myriad-Dreamin:
looks pretty good, but we might revert changes in changelogs.

Andrew15-5:
Why? The changes in the changelog didn't change, only the wording did.

Myriad-Dreamin:
> Why? The changes in the changelog didn't change, only the wording did.

Well, it is just less meaningful for me, because people usually don't read old changelogs. I read https://github.com/olivierlacan/keep-a-changelog and didn't find an answer.

Myriad-Dreamin (approved):

Myriad-Dreamin (approved):

Andrew15-5:
I think consistency is still pretty important, and you probably didn't find an answer because patching a changelog file doesn't mean anything bad, at least if the changes weren't altered so that the meaning is different. This is why I think it's totally fine, it's just not common enough, I guess.

PR #1521, opened 2025-03-16
feat: try setting root_dir to TYPST_ROOT in Neovim

Andrew15-5:
Answering [my own question](https://discord.com/channels/1054443721975922748/1260973804637786224/1350083276496441444). Since my project structure is somewhat complex and I do use `TYPST_ROOT` for some of its parts, it's nice that tinymist can finally follow it automatically. Now no absolute path will show an error reading a file.

### DIFF
--- a/editors/neovim/plugins/tinymist.lua
+++ b/editors/neovim/plugins/tinymist.lua
@@ -31,7 +31,7 @@ return {
             return vim.fn.getcwd()
           end
         end,
-        --- See [Tinymist Server Configuration](https://github.com/Myriad-Dreamin/tinymist/blob/main/Configuration.md) for references.
+        --- See [Tinymist Server Configuration](https://github.com/Myriad-Dreamin/tinymist/blob/main/editors/neovim/Configuration.md) for references.
         settings = {},
       }
     end,
--- a/editors/neovim/plugins/tinymist.lua
+++ b/editors/neovim/plugins/tinymist.lua
@@ -26,10 +26,10 @@ return {
         single_file_support = true,
         root_dir = function()
           if vim.env.TYPST_ROOT ~= nil then
-            return Path:new(vim.env.TYPST_ROOT):absolute()
-          else
-            return vim.fn.getcwd()
+            local typst_root = Path:new(vim.env.TYPST_ROOT)
+            if typst_root:exists() then return typst_root:absolute() end
           end
+          return vim.fn.getcwd()
         end,
         --- See [Tinymist Server Configuration](https://github.com/Myriad-Dreamin/tinymist/blob/main/editors/neovim/Configuration.md) for references.
         settings = {},
--- a/editors/neovim/plugins/tinymist.lua
+++ b/editors/neovim/plugins/tinymist.lua
@@ -1,3 +1,4 @@
+---@type LazySpec[]
 return {
   -- requires tinymist
   {
@@ -13,22 +14,26 @@ return {
     "neovim/nvim-lspconfig",
     dependencies = {
       "mason.nvim",
+      "nvim-lua/plenary.nvim",
       "williamboman/mason-lspconfig.nvim",
     },
-    ---@class PluginLspOpts
-    opts = {
-      ---@type lspconfig.options
-      servers = {
-        tinymist = {
-          --- todo: these configuration from lspconfig maybe broken
-          single_file_support = true,
-          root_dir = function()
+    config = function()
+      local lspconfig = require "lspconfig"
+      local Path = require "plenary.path"
+
+      lspconfig.tinymist.setup {
+        --- todo: these configuration from lspconfig maybe broken
+        single_file_support = true,
+        root_dir = function()
+          if vim.env.TYPST_ROOT ~= nil then
+            return Path:new(vim.env.TYPST_ROOT):absolute()
+          else
             return vim.fn.getcwd()
-          end,
-          --- See [Tinymist Server Configuration](https://github.com/Myriad-Dreamin/tinymist/blob/main/Configuration.md) for references.
-          settings = {}
-        },
-      },
-    },
+          end
+        end,
+        --- See [Tinymist Server Configuration](https://github.com/Myriad-Dreamin/tinymist/blob/main/Configuration.md) for references.
+        settings = {},
+      }
+    end,
   },
 }


Andrew15-5:
I see that the Typst-based docs is cooking, but there are also many Markdown files where related stuff are documented, so I don't want to mess with everything. I will wait until it's all unified, and maybe then I can add a few patches here and there. Or should Neovim users unify their docs? Right now it feels suboptimal.

Andrew15-5:
I was also thinking that maybe this `root_dir` setup can be upstreamed, but since it uses plenary, I don't think it will work.

Myriad-Dreamin:
I don't see obvious issue to set it at lua side. I have no knowledge to tell whether it is a good practice and nvim-lspconfig will accept it.

> I see that the Typst-based docs is cooking, but there are also many Markdown files where related stuff are documented, so I don't want to mess with everything. I will wait until it's all unified, and maybe then I can add a few patches here and there. Or should Neovim users unify their docs? Right now it feels suboptimal.

the markdown files are generated so you can only edit typst files. I can help generate them for you.

Andrew15-5:
> I don't see obvious issue to set it at lua side.

I don't understand why you say this.

> the markdown files are generated so you can only edit typst files. I can help generate them for you.

For some reason, `docs:typ` is broken. I had to change root from `"../.."` to ".."` and also it uses `target/debug/typlite` instead of globally installed one. And also the globally installed one doesn't know what `strike` variable is. But after building it from scratch, the command finishes successfully.

But it actually doesn't touch the Lua file, so I don't see the point of updating docs in this PR. Otherwise, the Lua content should be directly included into Typst docs.

Myriad-Dreamin:
> I don't understand why you say this.

You can also make tinymist-cli respect `TYPST_ROOT`. This deserves some discussion.

editors/neovim/plugins/tinymist.lua
```diff
@@ -13,22 +14,26 @@ return {
...
             return vim.fn.getcwd()
-          end,
-          --- See [Tinymist Server Configuration](https://github.com/Myriad-Dreamin/tinymist/blob/main/Configuration.md) for references.
-          settings = {}
-        },
-      },
-    },
+          end
+        end,
+        --- See [Tinymist Server Configuration](https://github.com/Myriad-Dreamin/tinymist/blob/main/Configuration.md) for references.
```
Myriad-Dreamin, line 34: A broken link is found. It is not related to this PR.

Andrew15-5:
> You can also make tinymist-cli respect `TYPST_ROOT`. This deserves some discussion.

Oh, well, I asked about this almost 3 days ago with no reply, so I then yesterday had an idea how to do this in Lua, so I made a PR to share the solution.

You mean `tinymist` CLI? I actually don't know about this for sure. In theory, the env var is exported and is passed to Neovim, which I would think is then passed to tinymist. If that's true, then you're probably right. I think the logic would be the same: if the environment variable is set, then we use it, if not, then use the root dir passed from the client.

Myriad-Dreamin:
> In theory, the env var is exported and is passed to Neovim, which I would think is then passed to tinymist. If that's true, then you're probably right. I think the logic would be the same: if the environment variable is set, then we use it, if not, then use the root dir passed from the client.

Correct. If we synchronize code from typst-cli everything will work (if, luckily). I'll thinking about the case a unskilled user occasionally makes a broken `TYPST_ROOT` and they complains about it. But generally it is great to follow all behavior of typst-cli unless it is impossible or harmful, otherwise it is a bug of tinymist.

https://github.com/Myriad-Dreamin/tinymist/blob/b2e9d7bbcd61ca9bf38b0eff522572b9e772c638/crates/tinymist-world/src/args.rs#L60-L62

Andrew15-5:
I mean, if it so happens that the root dir is indeed broken, then tinymist can always fall back to the PWD.

Myriad-Dreamin:
> I mean, if it so happens that the root dir is indeed broken, then tinymist can always fall back to the PWD.

I think it shouldn't use PWD, because a language server is not ensured to start with current directory to `PWD`. vscode sets it to the extension folder `$HOME/.vscode/extensions/tinymist`.

editors/neovim/plugins/tinymist.lua
```diff
@@ -13,22 +14,26 @@ return {
...
-          root_dir = function()
+    config = function()
+      local lspconfig = require "lspconfig"
+      local Path = require "plenary.path"
+
+      lspconfig.tinymist.setup {
+        --- todo: these configuration from lspconfig maybe broken
+        single_file_support = true,
+        root_dir = function()
+          if vim.env.TYPST_ROOT ~= nil then
```
Myriad-Dreamin, line 28: when `TYPST_ROOT` can be bad, it is probably not a good default settings. For example, `TYPST_ROOT` is set to `/home/a/work` but we will edit `/home/b/work/main.typ`.

editors/neovim/plugins/tinymist.lua
```diff
@@ -13,22 +14,26 @@ return {
...
-    opts = {
-      ---@type lspconfig.options
-      servers = {
-        tinymist = {
-          --- todo: these configuration from lspconfig maybe broken
-          single_file_support = true,
-          root_dir = function()
+    config = function()
+      local lspconfig = require "lspconfig"
+      local Path = require "plenary.path"
```
Myriad-Dreamin, line 22: Is it an external dependency? `plenary.path` shouldn't be used in default setting.

Andrew15-5:
> I think it shouldn't use PWD, because a language server is not ensured to start with current directory to `PWD`.

Uhh, then why do you use current dir?

https://github.com/Myriad-Dreamin/tinymist/blob/47410c89e88e8d09c33bbcc7d68735ea558aff1b/crates/tinymist/src/main.rs?plain=1#L158-L162

editors/neovim/plugins/tinymist.lua
Andrew15-5: Yes. Who says it shouldn't? It's probably not allowed in the nvim-lspconfig repo PR, but for user's config it's 100% valid.

This is not a default setting, it's user's custom setting.

editors/neovim/plugins/tinymist.lua
Andrew15-5: Fixed.

Myriad-Dreamin:
> > I think it shouldn't use PWD, because a language server is not ensured to start with current directory to `PWD`.
> 
> Uhh, then why do you use current dir?
> 
> https://github.com/Myriad-Dreamin/tinymist/blob/47410c89e88e8d09c33bbcc7d68735ea558aff1b/crates/tinymist/src/main.rs?plain=1#L158-L162

That's not the `lsp` command.

editors/neovim/plugins/tinymist.lua
Myriad-Dreamin: > Yes. Who says it shouldn't? It's probably not allowed in the nvim-lspconfig repo PR, but for user's config it's 100% valid.

How do we ensure that it is installed? Is it a dependency of lazyVim? Or is it automatically installed if it is missing? 

> This is not a default setting, it's user's custom setting.

If the plugin is placed here, it should be a default setting as well as an example and work for everyone. I should provide an example using `lspconfig.setup` but I'm not a neovim expert so I only put a worked settings used by my local LazyVim.

Myriad-Dreamin: In fact `lspconfig.setup` is perhaps also not suitable for all neovim distros. I cannot review details since I lack of concrete knowledge. @Eric-Song-Nop @max397574 @SylvanFranklin 

@SylvanFranklin You deleted the link to the example file in the [docs: Configuration](https://github.com/Myriad-Dreamin/tinymist/tree/1b5d4ac786f76d2317ee09f02d847ab1e9ceec22/editors/neovim#configuration).

Andrew15-5: > How do we ensure that it is installed? Is it a dependency of lazyVim? Or is it automatically installed if it is missing?

https://github.com/Myriad-Dreamin/tinymist/pull/1521/files#diff-541ead0f670156e6c30430b0ee62b582d9e28384ca0834bd0ca36bf9bc55e12dR17

Andrew15-5: > In fact `lspconfig.setup` is perhaps also not suitable for all neovim distros.

If someone doesn't use nvim-lspconfig, then they are probably rawdogging their whole setup and don't need any dependency plugin. Most people use it. I haven't looked into other distros, but I'm fairly confident all of them use it too.

Myriad-Dreamin (approved):

Eric-Song-Nop:
I don't like this idea, this might miss lead some users that tinymist for some tools will set or use the env var TYPST_ROOT, while it is not really a common use.

max397574:
I wouldn't add this like this, if at all. 1. I'd try working with the root_patterns and upstream this to lspconfig and 2. if for some reason we can't do this I'd do it without plenary which gets used less and less these days

Andrew15-5:
> I don't like this idea, this might miss lead some users that tinymist for some tools will set or use the env var TYPST_ROOT, while it is not really a common use

How is this misleading? After watching [a video](https://www.youtube.com/watch?v=5Welk51oDWs), I more clearly see that indeed, the average Neovim enjoyer is a person that likes tinkering and make their editor look and feel exactly how they want. It's clearly written that the environment is checked manually, there is absolutely nothing misleading about this.

> Neovim users should be able to find their prefered way of setting the project root

Well, at least now they would have _a_ way to do this, it's not particularly easy or nice to figure out everything by yourself. Although the logic is pretty basic, I just didn't have to use the variable before, and now that I do, and it's definitely not the same the .git parent dir, the idea appeared to me after a few annoying cases where the paths were not recognized.

The problem with `.git` and other is that it's not always where the root is, and the same goes to `main.typ` (and the name can be different). But if `TYPST_ROOT` is set, then it's very clear that this should be the root, because that is what `typst` would use, then it's as clear as day that tinymist should use it as well (if it's set). So in this regard, the proposed default is better than it is right now.

> I'd try working with the root_patterns and upstream this to lspconfig 

What exactly do you want to upstream? IIUC, `util.root_pattern` is the same as `vim.fs.root`.

> if for some reason we can't do this I'd do it without plenary which gets used less and less these days

If you are saying that I have to re-implement `absolute` method and all the normalization, then it's a waste of time and plenary handles this perfectly fine. I don't use plenary if I can get around it (I only used it maybe once or twice and only for paths stuff), which the modern API helps with, but if I can't, then it's a perfectly valid tool to use, with the exception that you have to dig into the codebase instead of readying non-existent documentation (I guess you can _kinda_ use Python docs for this).

---

I will no longer defend this PR, as it clearly doesn't have high chances to be merged, and I just wanted to share what I've made anyway. The only other thing I can offer is that we can provide 2–3 examples of how you _can_ set up `root_dir` to fit your needs. Then the reader can choose to use one of them, or none at all.

Myriad-Dreamin:
I'm not quite opposed to add `TYPST_ROOT` to tinymist, but we should have a clear view on how tinymist as a language server resolves root directories for typst files. Currently it is undocumented but pretty simple. There is a list in order of priority:

1. if a `tinymist.root` is passed by configuration, it is used in highest priority.
1. if a `tinymist.typstExtraArgs = ["--root", someRoot]` is passed by configuration, `someRoot` is used in then.
1. if `rootUri`, `rootPath`, or `workspaceFolders` is passed when lsp request initializations, it is used then in order.
    - `workspaceFolders` is used first, as lsp claimed. If there are multiple nested roots in the array, the first root covering the source file wins.
    - `rootUri` is used then, as lsp claimed.
    - `rootPath` is used in the end, as lsp claimed.
1. if a typst source file is located outside all of above root directories in `1`, `2` or `3`, the parent directory of the source file is considered as the root directory to use.
1. if tinymist has no permission to access all of above directories, or the source file has no root (for example, a memory buffer which doesn't have a file path), it will serve the requests on the source file with rootless mode.

If we add `TYPST_ROOT` detection when initializing the langauge server, it will use `TYPST_ROOT` with slightly lower priority than the `2`. That is it will have lower priority than the root passed by `tinymist.typstExtraArgs` and higher priority than roots negotiated in lsp standard initialization.

Myriad-Dreamin:
It is probably not good to detect `TYPST_ROOT` in neovim plugin. Another PR to add `TYPST_ROOT` to `tinymist.typstExtraArgs` is possible, but we should also add docs and tests at the same time.

PR #1539, opened 2025-03-19
feat: add diagnostics refiner to provide extra hints from tinymist side

### LINKED ISSUE #1428: Hint to use tinymist v0.12 when a document is broken in tinymist (typst) v0.13
**Motivation**

I receive many false bug reports everyday, where the error is caused by their templates are incompatible with typst v0.13.

- https://github.com/Myriad-Dreamin/tinymist/issues/1426

**Description**

I understand most people don't really know typst scripting and template things, while vscode implicitly upgrades tinymist to v0.13.0 and cause errors in their template. In this case, it is great to detect errors and guide them to solve the errors.

**Examples/Questions**

When the diagnostic message is "unknown argument: style", we attach a hint to downgrade tinymist to v0.12.x to use typst v0.12. This can solve their issues.

This mechanism can be used for automatically solving more compile errors. This will work like rust's clippy.

seven-mile:
Fix #1428. To refine the diagnostics from Typst compiler, this PR adds a shallow filter in `convert_diagnostics`, which originally converts `TypstDiag` to tinymist LSP `Diag`.

`Cow<_>` is used to avoid any unwanted copy when no refinement is needed.

This patch only contains the hint mentioned in #1428, several more patterns should come soon. A new E2E test is attached.


crates/tinymist-query/src/diagnostics.rs
```diff
@@ -168,3 +186,25 @@ fn diagnostic_hints(typst_hints: &[EcoString]) -> Format<impl Iterator<Item = Ec
...
+    fn matches(&self, raw: &TypstDiagnostic) -> bool {
+        raw.message.contains("unknown variable: style")
+    }
+
+    fn refine(&self, raw: TypstDiagnostic) -> TypstDiagnostic {
+        raw.with_hint(concat!(
+            r#"Typst 0.13 has introduced breaking changes. Try downgrading "#,
+            r#"Tinymist to v0.12 to use a compatible version of Typst, "#,
+            r#"or consider migrating your code according to "#,
+            r#"[this guide](https://typst.app/docs/changelog/0.13.0/#deprecations)."#
```
seven-mile, line 207: The new LSP spec added a client capability of presenting diagnostic messages in Markdown format, but currently no clients have implemented it yet, including VSCode. So we will see raw strings in IDE for now.

editors/vscode/src/test/e2e/simple-docs.test.ts
```diff
@@ -107,6 +107,19 @@ export async function getTests(ctx: Context) {
       await vscode.commands.executeCommand("workbench.action.closeActiveEditor");
     });
 
+    suite.addTest("typst0.13 diag hints", async () => {
```
Myriad-Dreamin, line 110: We are adding some expert testing in the simple docs test site. split both diagnostics cases to a new directory could be great.

crates/tinymist-query/src/diagnostics.rs
Myriad-Dreamin: For it is easy to render markdown markups in brain, this is not a big problem. We'll allow markups in diag messages.

crates/tinymist-query/src/diagnostics.rs
```diff
@@ -168,3 +186,25 @@ fn diagnostic_hints(typst_hints: &[EcoString]) -> Format<impl Iterator<Item = Ec
         .interleave(typst_hints.iter().cloned())
         .format("")
 }
+
+trait DiagnosticRefiner {
+    fn matches(&self, raw: &TypstDiagnostic) -> bool;
+    fn refine(&self, raw: TypstDiagnostic) -> TypstDiagnostic;
+}
+
+struct Typst0_13DeprecationRefiner();
```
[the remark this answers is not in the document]
Myriad-Dreamin: `Typst` is redundant. `0_13` can be a const argument, then it would be:

```rs
struct DeprecationRefiner<const MINOR: usize> {}

impl<MINOR> DeprecationRefiner {
  fn refine(&self, raw: TypstDiagnostic) -> TypstDiagnostic {
        raw.with_hint(format!("Typst 0.{MINOR} has introduced breaking changes."))
  }
}

impl DeprecationRefiner<13> {
    fn matches(&self, raw: &TypstDiagnostic) -> bool {
        raw.message.contains("unknown variable: style")
    }
}
```

crates/tinymist-query/src/diagnostics.rs
```diff
@@ -57,7 +60,21 @@ fn convert_diagnostic(
...
     typst_diagnostic: &TypstDiagnostic,
 ) -> anyhow::Result<(Url, Diagnostic)> {
-    let (id, span) = diagnostic_span_id(ctx, typst_diagnostic);
+    let typst_diagnostic = {
+        let mut diag = Cow::Borrowed(typst_diagnostic);
+
+        // Extend more refiners here by adding their instances.
+        let refiners = &[&Typst0_13DeprecationRefiner {}];
+
+        for refiner in refiners {
```
Myriad-Dreamin, line 69: There could be optimization room to caching the conversion, but I don't know the how delicious the benefit is.

seven-mile: Added a note for optimization potential.

crates/tinymist-query/src/diagnostics.rs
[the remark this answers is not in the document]
seven-mile: Naming suggestion applied. It seems rust does not allow partial `impl` for a trait, so `impl` of `refine` method remains separate for every minor version rather than a common one.

editors/vscode/src/test/e2e/simple-docs.test.ts
seven-mile: Test structure reorganized.

crates/tinymist-query/src/diagnostics.rs
Myriad-Dreamin: Thought of it a bit. At most time, the deprecated messages are not show up in nice documents.

Myriad-Dreamin (approved):

PR #1544, opened 2025-03-19
feat: fledge diagnostic hint patterns by updating 2 refiners

### LINKED ISSUE #117: More friendly "out of root hints"
The current hint is actually specific for `typst-cli`.

image

### LINKED ISSUE #1428: Hint to use tinymist v0.12 when a document is broken in tinymist (typst) v0.13

seven-mile:
This PR covers

* More hints of downgrading for diagnostics caused by feature removed in Typst v0.13 that are not properly hinted by Typst compiler already.
  * Should fix #1428 completely.
* Refined hints about "out of root imports" to make it not that confusing to IDE users.
  * Basically we removed the hint about change project root by appending a CLI argument `--root`.
  * Fix #117 .

The E2E test `typst013.typ` is extended to a multi-case test. A simple regex-based parser is applied to split it into sections.

A new E2E test `out-of-root.typ` is attached with a so-called `out-of-root/library.typ` for its importing.


crates/tinymist-query/src/diagnostics.rs
```diff
@@ -197,7 +198,15 @@ struct DeprecationRefiner<const MINOR: usize>();
 
 impl DiagnosticRefiner for DeprecationRefiner<13> {
     fn matches(&self, raw: &TypstDiagnostic) -> bool {
-        raw.message.contains("unknown variable: style")
+        let msg = &raw.message;
+        let patterns = [
```
Myriad-Dreamin, line 202: Can be a regex to perform $O(n)$ search, see `RegexSet`.

Myriad-Dreamin (approved):

ParaN3xus:
Maybe also
- block may not occur inside of a paragraph and was ignored
- unknown symbol modifier (for removed symbols like `degree.c`)
- cannot spread content (`..` in math equations)

seven-mile:
@ParaN3xus Thanks for adding!

> block may not occur inside of a paragraph and was ignored

After a rough check, it seems this one is currently a warning in v0.13.1. Probably we could add it when it's causing non-obvious migration debt for common users.

> unknown symbol modifier (for removed symbols like `degree.c`)

That would be nice! But we probably need a precise pattern matching to avoid provide confusing hints on false positives. @Myriad-Dreamin recommends implement a code action for symbol migration mapping, which involves much more powerful analysis infrastructure (compared with this string-based diag pattern matcher) and even better user experience. Wdyt?

> cannot spread content (`..` in math equations)

Nice catch! Suggestion applied.

Myriad-Dreamin:
For the "cannot spread content" pattern, it seems to have high false positive with only matching the error message but we are always hinting a downgrade suggestion.

PR #1575, opened 2025-03-24
feat: eject preview panel to browser

### LINKED ISSUE #953: Eject button on preview panel
**Motivation**

People usually only know the trick that run typst-preview in browser to avoid laggy editing in vscode via mouth of developers and regular users. This button as an entrance can help discover the feature.

**Description**

A "eject panel" button could help eject the panel quickly when experiencing lag. Additional trick tip can be attach to this button as well.

seven-mile:
This PR adds an "eject" button to the typst-preview webview panel. After clicking it, the preview switches its kind from `webview` to `browser`, overcoming some performance limitation caused by VSCode webview.

The implementation is specific to Preview-LSP. A global state `activePreviewPanelContext` is introduced and maintained during the lifecycle of every webview panel, since VSCode extension API does not allow us to detect the active panel.

Fix #953.

Locally tested manually by switching between two opening preview panels. Involving with external browser makes it hard to be exercised by a testcase. Please leave your idea if a proper test is required for this patch.


seven-mile:
This is how does it looks like:

image

editors/vscode/src/features/preview.ts
```diff
@@ -32,6 +33,14 @@ import { IContext } from "../context";
  */
 let launchImpl: typeof launchPreviewLsp;
 
+/**
+ * The task corresponding to the active preview panel.
+ */
+let activePreviewPanelContext: {
```
Myriad-Dreamin, line 39: The global state variable could be moved to `state.ts`. You can still define the interface of the context in `preview.ts`. It could be refactored like that:

```ts
// preview.ts
interface PreviewPanelContext {}
// state.ts
import type PreviewPanelContext from 'preview.ts'
interface ExtensionState {
  mut: {
    preview: PreviewPanelContext 
  }
}
```

editors/vscode/src/features/preview.ts
```diff
@@ -253,20 +304,38 @@ export async function openPreviewInWebView({
...
+    }
+  };
+
+  // NOTE: To avoid missing the auto revealing of webview initialization.
+  updateActivePanel();
+  panel.onDidChangeViewState(updateActivePanel);
+
   // todo: bind Document.onDidDispose, but we did not find a similar way.
   panel.onDidDispose(async () => {
+    if (activePreviewPanelContext?.panel === panel) {
```
Myriad-Dreamin, line 330: while there will be multiple panels, but we only keep a single `activePreviewPanelContext`. I'm okay if we don't consider multiple panels in this PR.

Myriad-Dreamin (approved):

editors/vscode/src/features/preview.ts
seven-mile: Refactor applied.

editors/vscode/src/features/preview.ts
seven-mile: Yes, "eject" is designed to be a per-panel action. This is renamed to `focusingPreviewPanelContext` (which also maintains the alignment with other fields in `state.ts`).

editors/vscode/src/features/preview.ts
```diff
@@ -754,6 +823,7 @@ interface PersistPreviewState {
   mode: "doc" | "slide";
   isNotPrimary: boolean;
   isBrowsing: boolean;
+  isDev: boolean;
```
seven-mile, line 828: Note that I also added the missing `isDev` field to `PersistPreviewState` in this patch, which affects serialization of webviews.

editors/vscode/src/features/preview.ts
```diff
@@ -173,6 +184,48 @@ export function previewActivate(context: vscode.ExtensionContext, isCompat: bool
...
+      vscode.window.showWarningMessage("No active preview panel");
+      return;
+    }
+    const { panel, state } = focusingContext;
+
+    // Close the preview panel, basically kill the previous preview task.
+    panel.dispose();
+
+    await launchForURI(vscode.Uri.parse(state.uri), "browser", state.mode, {
+      isBrowsing: state.isBrowsing,
```
Myriad-Dreamin, line 225: Are we missing a `isNotPrimary` here? The "primary" means that it is compiled by the primary compiler, see `ProjectCompiler`, usually the first previewed document.

editors/vscode/src/features/preview.ts
```diff
@@ -173,6 +184,48 @@ export function previewActivate(context: vscode.ExtensionContext, isCompat: bool
...
+  async function ejectPreviewPanelLsp() {
+    const focusingContext = extensionState.getFocusingPreviewPanelContext();
+    if (!focusingContext) {
+      vscode.window.showWarningMessage("No active preview panel");
+      return;
+    }
+    const { panel, state } = focusingContext;
+
+    // Close the preview panel, basically kill the previous preview task.
+    panel.dispose();
```
Myriad-Dreamin, line 222: The previous preview task in the srver is requested to destruct asynchonrously here, but we are not waiting them. Not sure whether it is a problem.

PR #1589, opened 2025-03-26
fix: correct `jump_from_cursor` and add tests

Myriad-Dreamin:
Since there was some stupid refactor, I also do some simple tests to examine it works on at least some simple documents.

### DIFF
--- a/crates/tinymist-query/src/jump.rs
+++ b/crates/tinymist-query/src/jump.rs
@@ -103,9 +103,10 @@ fn jump_from_cursor_(
     source: &Source,
     cursor: usize,
 ) -> Option<Vec<Position>> {
-    let node = LinkedNode::new(source.root())
-        .leaf_at_compat(cursor)
-        .filter(|node| !matches!(node.kind(), SyntaxKind::Text | SyntaxKind::MathText))?;
+    let node = LinkedNode::new(source.root()).leaf_at_compat(cursor)?;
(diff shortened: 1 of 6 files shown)


crates/tinymist-query/src/jump.rs
```diff
@@ -104,31 +104,43 @@ fn jump_from_cursor_(
     cursor: usize,
 ) -> Option<Vec<Position>> {
     let node = LinkedNode::new(source.root()).leaf_at_compat(cursor)?;
+    // todo: When we click on a label or some math operators, we seems likely also
```
michaelfortunato, line 111: Yes I agree with this point, could we not just remove the filter?

Myriad-Dreamin: It seems not good to comletely remove the filter, when we click on some `#let` binding nodes, which don't occur in the final compiled document, we might not expect that it will make preview panel scroll to some random position.

michaelfortunato: That is a good counter-example to my suggestion. Thank you. In that case, I would expect we would find the nearest renderable content to the `#let` binding and move the pdf there. 

For instance, if
```typst
= h1

#let x = [ Hi ]

Some text
```  
we would jump to the location corresponding to either =h1 or "Some text" in the pdf.

But this might not be a natural behavior

PR #1598, opened 2025-03-28
feat: fold list and enum item content

### LINKED ISSUE #1265: Allow folding lists
## Feature Request
**Motivation**

A section in a list, (numbered or not), contains everything that's in the right indent level below it, just like python blocks.

It would allow, for example, folding all levels, to reduce clutter, or navigate more easily.

```
+ Entry 1

  ...
   content
  ...
+ Entry 2

   ...
   content2
   ...
``` 

Since this content can be very long, it'd be useful if we can fold it, like we fold blocks in a programming language.

**Description**

Allow folding of list entries, like a programming languages's blocks. 

**Examples/Questions**

In this example we can fold the `#proof[...]` block, and the math block. I want to be able to also fold the numbered list entries. 

Image

It should probably leave the first line as is, just like you'd leave the `if ...:` but fold the block underneath in python.

BlueQuantumx:
An attempt to fix #1265. Currently it can only fold the content of an list/enum item, not a sequence of adjacent items.

Actually I'm not really sure about these changes, please check if there's any problems!

### DIFF
--- a/crates/tinymist-query/src/fixtures/folding_range/block-in-list-enum.typ
+++ b/crates/tinymist-query/src/fixtures/folding_range/block-in-list-enum.typ
@@ -1,0 +1,27 @@
+- Hi there
+
+  #[
+    This is a content block.
+  ]
+
+  #{
+    "This is a code block."
+  }
+
+  $
+    "This is a math block."
+  $
+
++ Hi there
+
+  #[
+    This is a content block.
+  ]
+
+  #{
+    "This is a code block."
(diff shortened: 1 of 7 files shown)


Myriad-Dreamin:
I tested it a bit and it seems work well. We should only add more tests like nested list, nested enum, or block in list to catch potential bugs.

BlueQuantumx:
I've just added some new tests, and fixed the format mistake. Please check.

PR #1611, opened 2025-04-01
feat: show rendered bibliography in bib hover and improve label hover

QuadnucYard:
The bib detail is displayed as in the bibliograph section. Locales specified for the citation elem are not considered for easy implementation.
The citation style follows that of the only bib elem.
This may help identify some issues in the csl. 

image
image

And wrap the figure value in a raw.

image


crates/tinymist-query/src/analysis/bib.rs
```diff
@@ -1,7 +1,26 @@
...
+    foundations::{Bytes, IntoValue, Packed},
+    introspection::Introspector,
+    model::{BibliographyElem, CslStyle},
+};
 use yaml_rust2::{parser::Event, parser::MarkedEventReceiver, scanner::Marker};
 
-use super::prelude::*;
+use super::{prelude::*, SharedContext};
+
+pub(crate) fn get_bib_elem_and_info(
```
Myriad-Dreamin, line 10: This and `render_citation_string` shouldn't be put into analysis crates, either leave them as private functions in the `hover` crate or create a `src/bib/mod.rs` to store bib specific functions.

Myriad-Dreamin (approved):

PR #1636, opened 2025-04-08
Add `clone_and_rebuild` so we can append fonts without a mutable access.

c0per:
Since `CompilerUniverse` only holds a `Arc<FontResolver>`, we can't call `append_fonts` or `rebuild` on it.

I changed the signature on `append_fonts` and added a new method `clone_and_build`, so it's possible to build a new `FontResolverImpl` with more fonts. The new `FontResolverImpl` can then be swapped to the universe by calling `increment_revision` method.

### DIFF
--- a/crates/tinymist-world/src/font/resolver.rs
+++ b/crates/tinymist-world/src/font/resolver.rs
@@ -125,7 +125,7 @@ impl FontResolverImpl {
         }
     }
 
-    pub fn append_font(&mut self, info: FontInfo, slot: FontSlot) {
+    pub fn append_font(&self, info: FontInfo, slot: FontSlot) {
         let mut font_book = self.partial_book.lock().unwrap();
         font_book.push((None, info, slot));
     }
--- a/crates/tinymist-world/src/font/resolver.rs
+++ b/crates/tinymist-world/src/font/resolver.rs
@@ -137,7 +137,7 @@ impl FontResolverImpl {
         }
         partial_book.revision += 1;
 
-        let mut book = FontBook::default();
+        let mut new_book = FontBook::default();
 
         let mut font_changes = HashMap::new();
         let mut new_fonts = vec![];
@@ -148,32 +148,23 @@ impl FontResolverImpl {
                 new_fonts.push((info, slot));
             }
         }
-        partial_book.changes.clear();
         partial_book.partial_hit = false;
 
-        let mut font_slots = Vec::new();
-        font_slots.append(&mut self.fonts);
-        self.fonts.clear();
-
-        for (i, slot_ref) in font_slots.iter_mut().enumerate() {
-            let (info, slot) = if let Some((_, v)) = font_changes.remove_entry(&i) {
-                v
+        for (i, slot_ref) in self.fonts.iter_mut().enumerate() {
+            if let Some((info, slot)) = font_changes.remove(&i) {
+                new_book.push(info);
+                *slot_ref = slot;
             } else {
-                book.push(self.book.info(i).unwrap().clone());
-                continue;
-            };
-
-            book.push(info);
-            *slot_ref = slot;
+                new_book.push(self.book.info(i).unwrap().clone());
+            }
         }
 
-        for (info, slot) in new_fonts.drain(..) {
-            book.push(info);
-            font_slots.push(slot);
+        for (info, slot) in new_fonts {
+            new_book.push(info);
+            self.fonts.push(slot);
         }
 
-        self.book = LazyHash::new(book);
-        self.fonts = font_slots;
+        self.book = LazyHash::new(new_book);
     }
 
     pub fn add_glyph_packs(&mut self) {
--- a/crates/tinymist-world/src/font/resolver.rs
+++ b/crates/tinymist-world/src/font/resolver.rs
@@ -167,6 +167,45 @@ impl FontResolverImpl {
         self.book = LazyHash::new(new_book);
     }
 
+    pub fn clone_and_rebuild(&self) -> Self {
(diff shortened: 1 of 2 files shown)


Myriad-Dreamin:
The partial book is a pretty old and outdated API. I would suggest not using this because the order of font in the `World::font_book` is important while you might ignore the order when usingthe outdated API:

https://github.com/Myriad-Dreamin/tinymist/blob/01afa463f402eaafc5b6c61fffe2aed66b81283d/crates/tinymist-world/src/font/system.rs#L119-L137

Instead, we should build the `FontResolverImpl` from scratch and edit the universe using the `RevisingUniverse::set_fonts`:

https://github.com/Myriad-Dreamin/tinymist/blob/01afa463f402eaafc5b6c61fffe2aed66b81283d/crates/tinymist-world/src/world.rs#L378-L381

If the overhead of build from scratch is not neglible, we should improve the performance of `SystemFontSearcher` or other font searchers by some approach.

What's your opinion?

Myriad-Dreamin:
> I think we should remove partial_book altogether, since it's outdated and isn't really an optimization.

Looks good to me.

> I propose that, since we don't have mutable access to FontResolver in CompilerUniverse, self-edit methods doesn't really work.

I meant you shouldn't mutate a built font resolver. Instead you should get a new resolver by updating arguments. This is because one may happen to corrupt the order of fonts and typst cannot use fonts correctly.

```rs
// Initializes fonts.
let opts: CompileFontOpts = new_font_opts();
let fonts = SystemFontSearcher::new().resolve_opts(opts).build();

// now we can already do that.
let new_opts = update_font_opts(opts);
let new_fonts = SystemFontSearcher::new().resolve_opts(new_opts).build();

// I feel okay if we will reuse some font objects internally if this actually make rebuild fster.
let new_fonts = SystemFontSearcher::reuse(fonts).resolve_opts(new_opts).build();
```

The idea behind the `reuse` can be checked from `typst::utils::LazyHash::reuse`

https://github.com/typst/typst/blob/94a497a01ffd60743b0a2ae67367be168bbde076/crates/typst-utils/src/hash.rs#L54-L61C6

c0per:
I kind of get it.

So `SystemFontSearcher` and `BrowserFontSearcher` are builder types for a `FontResolver`. 
If we want to add or edit fonts in a resolver, we should `FontSearcher::reuse()` it to get a searcher (builder), then edit and rebuild.

By the way, what's the difference between `FontResolverImpl` and `TinymistFontResolver`? They seem to be the same.

Myriad-Dreamin:
> I kind of get it.
> 
> So `SystemFontSearcher` and `BrowserFontSearcher` are builder types for a `FontResolver`. If we want to add or edit fonts in a resolver, we should `FontSearcher::reuse()` it to get a searcher (builder), then edit and rebuild.

Great, exactly. We should put our discussion into doc comments to help people in future understand it. If you have more points about design or questions about the codebase, feel free to ask me.

> By the way, what's the difference between `FontResolverImpl` and `TinymistFontResolver`? They seem to be the same.

In 2024, `FontResolverImpl` and `TinymistFontResolver` was not in same repository. The `TinymistFontResolver` is a copy of the `FontResolverImpl` and used by `LspWorld`. Comparing with `FontResolverImpl`, `TinymistFontResolver` provides some introspecting API for IDE (editors). For example, given a font object, it tells the location the font was loaded from:

https://github.com/Myriad-Dreamin/tinymist/blob/71116758ef90934dce4feead6e677a8fb83a37f7/crates/tinymist-project/src/font.rs#L67-L68

With that, the editor can help check and debug font things:

image

It should be the time to merge them, but it stands on the long todolist on the refactor and document on the `tinymist_world`, so I haven't started it.

c0per:
Checkout [dd347a9](https://github.com/Myriad-Dreamin/tinymist/pull/1636/commits/dd347a9227c964489fa8b4025fac8cd0fa302327).

Since `typst::FontBook` doesn't provide any method to edit its content, we only can push more fonts into it, there's no need to maintain a FontBook in searcher. FontBook will only be built when it's converting to a resolver.

So font searcher is just a wrapper for `Vec<(FontInfo, FontSlot)>`, with multiple loading methods. A new method `with_fonts_mut()` provides mutable access to fonts directly.

What's the purpose of `FontProfile`? Couldn't find any usage of it.

Myriad-Dreamin:
> Since `typst::FontBook` doesn't provide any method to edit its content, we only can push more fonts into it, there's no need to maintain a FontBook in searcher. FontBook will only be built when it's converting to a resolver.
> 
> So font searcher is just a wrapper for `Vec<(FontInfo, FontSlot)>`, with multiple loading methods. A new method `with_fonts_mut()` provides mutable access to fonts directly.

Sounds good.

> What's the purpose of `FontProfile`? Couldn't find any usage of it.

The `FontProfile` is serializable to help cache `typst::FontInfo` into disk or send to remote server. The webapp (https://typst.app) also has their owned `FontProfile` to know which fonts are stored remotely. If we apply the cache, the time to load font will decrease from 120ms to 16ms on windows 10. But we are not doing that since it is hard to maintain them correctly, because, for example, there will be usually multiple tinymist processes and other tools to access a same font disk cache. We may finally to cache them when we have time to design it further.

Myriad-Dreamin:
To still allow to rebuild fonts, I try to add an `IncrFontSearcher`, which provides a mutable `ops` to edit the searcher. How do you think of it?

```rs
let mut searcher = IncrFontSearcher::default();

// Initial build
searcher.push_op(StaticAssets(vec![]));
searcher.push_op(SystemFonts);
searcher.build();

// Incremental build
let _ = searcher.set_op(0, StaticAssets(vec![]));
searcher.build();
```

PR #1649, opened 2025-04-11
feat(lint): add warning for vf font

Enter-tainer:
Please let me know if and how we can do better on this, for example:

1. resolve font info and see if it's vf or not(instead of check if it ends with "VF")
2. handle cases where fonts are not directly visible in ast

### DIFF
--- a/crates/tinymist-query/src/fixtures/lint/snaps/test@vf.typ.snap
+++ b/crates/tinymist-query/src/fixtures/lint/snaps/test@vf.typ.snap
@@ -1,0 +1,21 @@
+---
+source: crates/tinymist-query/src/analysis.rs
+expression: "JsonRepr::new_redacted(result, &REDACT_LOC)"
+input_file: crates/tinymist-query/src/fixtures/lint/vf.typ
+---
+{
+ "s0.typ": [
+  {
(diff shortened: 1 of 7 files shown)


crates/tinymist-query/src/fixtures/lint/snaps/test@vf.typ.snap
```diff
@@ -0,0 +1,15 @@
+---
+source: crates/tinymist-query/src/analysis.rs
+expression: "JsonRepr::new_redacted(result, &REDACT_LOC)"
+input_file: crates/tinymist-query/src/fixtures/lint/vf.typ
+---
+{
+ "s0.typ": [
+  {
+   "message": "variable font is not supported by typst yet\nHint: consider using a static font instead. For more information, see https://github.com/typst/typst/issues/185",
+   "range": "2:6:2:37",
```
Enter-tainer, line 10: todo: add check for set rule:

```
    SetRule: 30 [
        Set: "set",
        Space: " ",
        Ident: "text",
        Args: 22 [
            LeftParen: "(",
            Named: 20 [
                Ident: "font",
                Colon: ":",
                Space: " ",
                Str: "\"Noto Sans VF\"",
            ],
            RightParen: ")",
        ],
    ],
```

Myriad-Dreamin (approved):

PR #1655, opened 2025-04-14
feat: reasonable default values for settings for formatterMode

kaerbr:
I've added, for me, meaningful values for the settings `tinymist.formatterMode` and `tinymist.fontPaths`.

With this change it is possible to
- format `*.typ` without having to make changes to the settings (by default).
~~- use fonts, that are present in the `fonts` folder inside the workspace.~~

If you guys also think this setting makes sense, I would love to see this MR being merged. (:

### DIFF
--- a/editors/vscode/package.json
+++ b/editors/vscode/package.json
@@ -446,7 +446,7 @@
           "title": "%extension.tinymist.config.tinymist.formatterMode.title%",
           "markdownDescription": "%extension.tinymist.config.tinymist.formatterMode.desc%",
           "type": "string",
-          "default": "disable",
+          "default": "typstyle",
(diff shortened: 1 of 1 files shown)


Enter-tainer:
I'm okay with the formatter change. but i'm kinda disagree with the default font path. `./fonts` is not a well known path, and typst-cli doesnt recognize it.

kaerbr:
Ok. 
Then I will roll back the changes for `tinymist.fontPaths`, so that the default formatter can be merged. (:

Enter-tainer:
cc @Myriad-Dreamin what's your idea on default formatter? PTAL, thanks!

Issue #1595, opened 2025-03-28
Update neovim docs for deprecated function `vim.lsp.buf.execute_command`


TimeTravelPenguin:
### Motivation

As seen in this [deprication](https://neovim.io/doc/user/deprecated.html#vim.lsp.buf.execute_command), `vim.lsp.buf.execute_command` is to be replaced in favor of `client:exec_cmd`.

The tinymist docs do not currently reflect this change.

I am in the process of updating my config, but I am not entirely fluent in Lua, nor with neovim, so I am having some difficulty and cannot suggested a complete migration.

### Example

In my (lsp)config, I currently have
```lua
lspconfig.tinymist.setup {
  on_attach = function(client, bufnr)
    local map = vim.keymap.set

    map("n", "<leader>ba", function()
      client:exec_cmd({
        command = "tinymist.pinMain",
        arguments = { vim.api.nvim_buf_get_name(0) },
      }, { bufnr = bufnr })

      vim.print("Updated pinned main to " .. vim.api.nvim_buf_get_name(0))
    end, { desc = "tinymist: Pin buffer as main", noremap = true })
  end,
  -- other setup
}
```

I am not able to get the "unpinned" variant, as shown in the current docs, to work by setting `arguments = { nil }`.

Myriad-Dreamin:
From my memory, some people reported that `arguments = { nil }` doesn't work, but it does work on some other people's neovim, which is confusing me.

@Eric-Song-Nop @SylvanFranklin

TimeTravelPenguin:
I was previously using
```lua
map("n", "<leader>bd",
  function()
    vim.lsp.buf.execute_command { command = "tinymist.pinMain", arguments = { nil } }
  end, { desc = "tinymist: Unpin buffer as main", noremap = true })
```
and this was never a problem for me. But now, when I use the relevant format shown above, I get an error saying `tinymist: -32602: expect Option<PathBuf> at args[0]`.

SylvanFranklin:
What version of neovim are you running here?

Myriad-Dreamin:
> What version of neovim are you running here?

@TimeTravelPenguin 

> As seen in this [deprication](https://neovim.io/doc/user/deprecated.html#vim.lsp.buf.execute_command), `vim.lsp.buf.execute_command` is to be replaced in favor of `client:exec_cmd`.
> 
> The tinymist docs do not currently reflect this change.
> 
[...]

PR #1670, opened 2025-04-16
fix: Neovim unpin error `expect Option<PathBuf> at args[0]`

ptdewey:
Addresses #1595

### DIFF
--- a/crates/tinymist/src/cmd.rs
+++ b/crates/tinymist/src/cmd.rs
@@ -279,7 +279,11 @@ impl ServerState {
 
     /// Pin main file to some path.
     pub fn pin_document(&mut self, mut args: Vec<JsonValue>) -> AnySchedulableResponse {
-        let entry = get_arg!(args[0] as Option<PathBuf>).map(From::from);
+        let entry = if args.get(0).map_or(true, |v| v.is_null()) {
+            None
+        } else {
+            get_arg!(args[0] as Option<PathBuf>).map(From::from)
+        };
 
         let update_result = self.pin_main_file(entry.clone());
         update_result.map_err(|err| internal_error(format!("could not pin file: {err}")))?;


Myriad-Dreamin:
I don't thing it is a valid fix. From your fix, it is inferred that neovim ignores the `nil` argument, so the array passing to lsp is `[]` instead of `[null]`.

TimeTravelPenguin:
I am continuing the discussion from https://github.com/Myriad-Dreamin/tinymist/issues/1595#issuecomment-2811880712.

I am not very familiar with anything more than basic neovim/lua, or anything yet with tinymist's source code. But I am curious about the issue.

After taking a look at the proposed fix, while I don't understand the larger design of the codebase, my understanding is that the difficulty lies with the representation of the state of the pinned main file (or lack thereof).

@Myriad-Dreamin, you mention:
> From your fix, it is inferred that neovim ignores the nil argument, so the array passing to lsp is `[]` instead of `[null]`.

In the following section:
```rs
let entry = if args.first().is_none_or(|v| v.is_null()) {
    None
} else {
    get_arg!(args[0] as Option<PathBuf>).map(From::from)
};
```

Are you saying that the first branch here, which returns `None`, is effectively resulting in `[]` instead of `[null]`?

Myriad-Dreamin:
> I am continuing the discussion from [#1595 (comment)](https://github.com/Myriad-Dreamin/tinymist/issues/1595#issuecomment-2811880712).
> 
> I am not very familiar with anything more than basic neovim/lua, or anything yet with tinymist's source code. But I am curious about the issue.
> 
> After taking a look at the proposed fix, while I don't understand the larger design of the codebase, my understanding is that the difficulty lies with the representation of the state of the pinned main file (or lack thereof).
> 
> @Myriad-Dreamin, you mention:
> 
> > From your fix, it is inferred that neovim ignores the nil argument, so the array passing to lsp is `[]` instead of `[null]`.
> 
> In the following section:
> 
> ```rust
> let entry = if args.first().is_none_or(|v| v.is_null()) {
>     None
> } else {
>     get_arg!(args[0] as Option<PathBuf>).map(From::from)
> };
> ```
> 
> Are you saying that the first branch here, which returns `None`, is effectively resulting in `[]` instead of `[null]`?

Yes, to be specific, it is working in VS Code (passing `["path/string"]` and `[null]`) but not in neovim. It should work with the proposed fix, so I guess neovim is passing `[]` when we specify `arguments = { nil }`.

https://github.com/Myriad-Dreamin/tinymist/blob/fc5757a5bca683dd8ed52298c99af24841b2bcc2/editors/vscode/src/extension.ts#L380-L392

Please test it and correct me if I'm wrong.

ptdewey:
Lua interprets the table `{ nil }` as `{}`, which Tinymist does not accept as a valid input to the function. The underlying problem here seems to be that Lua's `nil` is treated as a non-value, which is not equivalent to JSON `null`.

I did however discover that Neovim/Vim has a variable for representing JSON null:
```lua
--- Special value used to put "null" in JSON and NIL in msgpack.
--- See `json_encode()`.  This value is converted to "v:null" when
--- used as a String (e.g. in `expr5` with string concatenation
--- operator) and to zero when used as a Number (e.g. in `expr5`
--- or `expr7` when used with numeric operators). Read-only.
--- In some places `v:null` can be used for a List, Dict, etc.
--- that is not set.  That is slightly different than an empty
--- List, Dict, etc.
--- @type vim.NIL
vim.v.null = ...
```

Updating my keybind with this does fix the issue:
```lua
        vim.keymap.set("n", "<leader>tu", function()
            client:exec_cmd({
                title = "unpin",
                command = "tinymist.pinMain",
                arguments = { vim.v.null }, -- Also works with `vim.NIL`
            }, { bufnr = bufnr })
            vim.print("Unpinned: " .. vim.api.nvim_buf_get_name(0))
        end, { desc = "[T]ypst [U]npin", noremap = true })
```

I am left feeling that this is a bit hacky, but if you don't want to change the underlying API that is your prerogative.

If we don't change the API, I can update this PR (or open a new one) updating the Neovim docs to fix the issue.

Myriad-Dreamin:
> I am left feeling that this is a bit hacky, but if you don't want to change the underlying API that is your prerogative.

Let's keep it to not just change to fix bug. Having to use `vim.NIL` instead of `nil` also looks like a quirk of lua to me, and we know there are quite a bit other ones in lua. I would like to consider changing it or not as we receiving more reports from other editors. For now, we only know the situations in neovim and vscode.

PR #1678, opened 2025-04-17
docs: Change pinning to use non-deprecated function `client:exec`

ptdewey:
Related to #1670 and #1595.

Should not be merged until #1670 is resolved.

### DIFF
--- a/docs/tinymist/frontend/neovim.typ
+++ b/docs/tinymist/frontend/neovim.typ
@@ -133,35 +133,9 @@ Tinymist cannot know the main file of a multiple-files project if you don't tell
 
 The solution is a bit internal, which should get further improvement, but you can pin a main file by command.
 
-Using `vim.lsp.config`:
 ```lua
-vim.lsp.config["tinymist"] = {
-    cmd = { "tinymist" },
-    filetypes = { "typst" },
-    settings = {}
-    on_attach = function(client, bufnr)
-        vim.keymap.set("n", "<leader>tp", function()
-            client:exec_cmd({
-                title = "pin",
-                command = "tinymist.pinMain",
-                arguments = { vim.api.nvim_buf_get_name(0) },
-            }, { bufnr = bufnr })
-        end, { desc = "[T]inymist [P]in", noremap = true })
-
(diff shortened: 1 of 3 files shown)


docs/tinymist/frontend/neovim.typ
```diff
@@ -133,13 +133,67 @@ Tinymist cannot know the main file of a multiple-files project if you don't tell
 
 The solution is a bit internal, which should get further improvement, but you can pin a main file by command.
 
+Using `vim.lsp.config`:
+```lua
+vim.lsp.config["tinymist"] = {
+    cmd = { "tinymist" },
+    filetypes = { "typst" },
+    settings = {}
+    on_attach = function(client, bufnr)
```
Myriad-Dreamin, line 142: It has same content as "Using `lspconfig`". Could we eliminate the duplication?

editors/neovim/plugins/tinymist.lua
```diff
@@ -24,7 +24,31 @@ return {
...
-          settings = {}
+          settings = {
+            --- You could set the formatter mode to use lsp-enhanced formatters.
+            -- formatterMode = "typstyle",
+            
+            --- If you love to edit the documents and preview exported pdfs in the same time,
+            --- you could set this to `onType`.
+            -- exportPdf = "onType",
+          },
+          on_attach = function(client, bufnr)
```
Myriad-Dreamin, line 35: I updated the `tinymist.lua` sample

docs/tinymist/frontend/neovim.typ
ptdewey: We likely could, but I think for clarity's sake, it is probably worth including both to avoid any confusion in terms of where the `on_attach` function should be included.

It should be fine to include only the `vim.lsp.config` version in the docs, and the `lspconfig` version in the `tinymist.lua` sample, maybe with a link to it.

ptdewey:
It may be worth adding another sample lua file that uses the builtin `vim.lsp.config` instead of `lspconfig`, but that should probably be in a different PR.

Myriad-Dreamin (approved):

Issue #1588, opened 2025-03-26
typlite is not evaulating enum markers properly


michaelfortunato:
I believe the code responsible for evaluating `+` is here: https://github.com/Myriad-Dreamin/tinymist/blob/ee1c0ace46d3b187558594195230fa17b8e3909f/crates/typlite/src/lib.rs#L814-L826

The evaluation function for `+` that actually increments the counter is here [`crates/typst-eval/src/markup.rs:#L254`. ](https://github.com/typst/typst/blob/1f1c1338785dc09a43292cf106b4a23b4e1bd86e/crates/typst-eval/src/markup.rs#L245)

To reproduce, feed typlite a document 
```typst
+ Hi 
+ Howdy
+ Hey
```
and observe that the markdown becomes 
```md
1. Hi
1. Howdy
1. Hey
```

New to the codebase, but I think typlite should _not_ cast the EnumItem as untyped on line 817.

https://github.com/Myriad-Dreamin/tinymist/blob/ee1c0ace46d3b187558594195230fa17b8e3909f/crates/typlite/src/lib.rs#L817

Basically, it needs to call eval for EnumItem properly, again see [`crates/typst-eval/src/markup.rs:#L254`. ](https://github.com/typst/typst/blob/1f1c1338785dc09a43292cf106b4a23b4e1bd86e/crates/typst-eval/src/markup.rs#L245)

I think what we need to do is something like the following in [`typlite/src/lib.rs:L814`](https://github.com/Myriad-Dreamin/tinymist/blob/ee1c0ace46d3b187558594195230fa17b8e3909f/crates/typlite/src/lib.rs#L814)

```rs
    fn enum_item(&mut self, node: &SyntaxNode) -> Result<Value> {
        let enum_item = node.cast::<ast::EnumItem>().unwrap();
        let val = enum_item.eval(self.world).unwrap();
        Ok(Value::Content(val.plain_text()))
    }
```
Though this does not compile because `self.world` is not what `eval` accepts. 

This might be a known limitation of typlite given the note on `context` to svg code in the README, in which case I apologize for raising the issue.

michaelfortunato:
I get the sense that we are not using World as defined in typst-eval in this code. I see that TypliteWorker has a bit of state attached to it, in particular the field,  list_depth, so perhaps we add enum_item_counter or the like to the struct to track the number. It's a bit ad-hoc but it works.

Myriad-Dreamin:
[...]

PR #1684, opened 2025-04-18
feat: reimplement typlite by html export

Myriad-Dreamin:
The implementation using HTML export, as mentioned in #1588, @michaelfortunato

Todolist:
- [x] fix docstring conversion cases.
- [ ] clean up code.


hongjr03:
elements created by the user calling `html.frame` inside that have predefined display rules applied to them may lead to unexpected results, which needs some further discussions i think?

for example:
```typ
#html.frame(
  grid(columns: 2, [hello], [world])
)
```
will render nothing in result.

hongjr03:
another issue should be concerned:
```typst
show figure: it => md-figure(it.body, caption: it.caption.body)
```
As caption can also be `content` type, what should we do to make this show rule work properly in all scene?

michaelfortunato:
Thanks for the mention. I will take a loo this weekend.

Myriad-Dreamin:
@michaelfortunato This PR needs some tidy up but I will merge it first, because it has been waiting for merge for a long time and the [later latex export and docx export](#1718) also relies on this. Overall, this PR implements:
- A `markdown.typ` contains and applies a bundle of show rules to make typst document suitable for further parsing.
- A parser (`typlite/src/parser`) that lowers the compiled HTML into AST. The AST resembles pandoc's IR, and can be converted to markdown, latex, docx and other markup. The definition of the AST is located in the cmark-writer crate.
- A writer (`typlite/src/writer/markdown.rs`) that converts a AST into markdown files.

PR #1696, opened 2025-04-21
feat: add path conversion actions for absolute and relative paths

BlueQuantumx:


Myriad-Dreamin:
I made some progress but then I didn't have time to make a review.
- I tried `construct_module_dependencies` but it doesn't quite help because no import nodes are recorded.
- I tried the `classify_syntax` from `matcher.rs` to match imports, which is also a syntax-only matcher.
- I edited `CodeActionWorker::work` so that it doesn't stop intemediately if a path rewrite action is found. The test `path_and_equation.typ` verified this.
- I added and passed tests `path_import.typ` and `path_json.typ`, but `path_import_paren.typ` was not passed.

BlueQuantumx:
> * I added and passed tests `path_import.typ` and `path_json.typ`, but `path_import_paren.typ` was not passed.

For the failure on `path_import_paren.typ`, I think it's intrinsically because Typst accepts an expression on the position of path. You can even write something like `#import "template/" + "template.typ"`. My opinion is just ignore that case, since nobody would write like that?

BlueQuantumx:
> > > * I added and passed tests `path_import.typ` and `path_json.typ`, but `path_import_paren.typ` was not passed.
> > 
> > 
> > For the failure on `path_import_paren.typ`, I think it's intrinsically because Typst accepts an expression on the position of path. You can even write something like `#import "template/" + "template.typ"`. My opinion is just ignore that case, since nobody would write like that?
> 
> It should be too complicated to handle `#import "template/" + "template.typ"`, but we can do better to at least handle the cases where only parentheses are wrapped around a simple path string. Netherless, we could add some tests, where imported expressions are not a simple string, to ensure that we don't generate wrong code actions on that expressions.

I've wrote a test `path_expression_import.typ`. But I'm still not in favor of resolving the "parenthesed" case. I think it's a discussion on the design principle. They are all expressions, if we don't deal with a complex expression, nor should we deal with a simple expression. I don't like introducing overheads for a totally avoidable corner case. I have no idea about in what situation people have to write things like `#import ("some/path.typ")`. If I make any mistake, I'd like to change my mind!

Myriad-Dreamin:
> I've wrote a test `path_expression_import.typ`. But I'm still not in favor of resolving the "parenthesed" case. I think it's a discussion on the design principle. They are all expressions, if we don't deal with a complex expression, nor should we deal with a simple expression. I don't like introducing overheads for a totally avoidable corner case. I have no idea about in what situation people have to write things like `#import ("some/path.typ")`. If I make any mistake, I'd like to change my mind!

`import(path)` is only a common misuse of the import syntax, where they think of `import` is a function, and I see it for many times on the discord server. This is a misuse but I think it is okay if they resist to use it. I'm okay to not support them, but I can see a future that some people report that the parenthesed case doesn't work, but we can only close it as no planed.

Myriad-Dreamin:
If you think of this is ready, you can undraft it. I'll merge it in v0.13.14 (next release).

BlueQuantumx:
> `import(path)` is only a common misuse of the import syntax, where they think of `import` is a function, and I see it for many times on the discord server. This is a misuse but I think it is okay if they resist to use it. I'm okay to not support them, but I can see a future that some people report that the parenthesed case doesn't work, but we can only close it as no planed.

Oh I see 👀. Maybe it would be a little opinionated, but I'd like to say let's merge it for now. Maybe we can improve it later.

Myriad-Dreamin (approved):

PR #1700, opened 2025-04-23
fix: correct parameter name in `diff` function for clarity

BlueQuantumx:
We generally call a relative path "from a base to a target".

The original code seems made a mistake in naming "to" and "from".


BlueQuantumx:
`path/base` is nice. I've changed them.

PR #1704, opened 2025-04-25
fix: built-in CSL styles are recognized as path link

BlueQuantumx:
Wrong case: 
image

### DIFF
--- a/crates/tinymist-query/src/analysis/link_expr.rs
+++ b/crates/tinymist-query/src/analysis/link_expr.rs
@@ -120,7 +120,7 @@ impl LinkStrWorker {
                     }
                     "bibliography" => {
                         self.analyze_reader(node, call, "cite", false);
-                        self.analyze_reader(node, call, "style", false);
+                        self.analyze_bibliography_style(node, call);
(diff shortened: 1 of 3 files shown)


crates/tinymist-query/src/analysis/link_expr.rs
```diff
@@ -134,6 +134,24 @@ impl LinkStrWorker {
         }
     }
 
+    fn analyze_bibliography_style(&mut self, node: &LinkedNode, call: ast::FuncCall) -> Option<()> {
+        for item in call.args().items() {
+            match item {
+                ast::Arg::Named(named) if named.name().get().as_str() == "style" => {
+                    if let ast::Expr::Str(style) = named.expr() {
+                        if hayagriva::archive::ArchivedStyle::by_name(&style.get()).is_some() {
```
Myriad-Dreamin, line 142: Can we resolve a link to these builtin styles? For example, `alphanumeric` is on https://github.com/typst/hayagriva/blob/9df6db230edd0f8f08dd11253c564426f5085156/styles/alphanumeric.csl

Myriad-Dreamin (approved):

PR #1718, opened 2025-04-30
feat: typlite supports latex export

Myriad-Dreamin:
There are unresolved questions from #1684:

> elements created by the user calling `html.frame` inside that have predefined display rules applied to them may lead to unexpected results, which needs some further discussions i think?
> 
> for example:
> 
> ```typst
> #html.frame(
>   grid(columns: 2, [hello], [world])
> )
> ```
> 
> will render nothing in result.

> another issue should be concerned:
> 
> ```typst
> show figure: it => md-figure(it.body, caption: it.caption.body)
> ```
> 
> As caption can also be `content` type, what should we do to make this show rule work properly in all scene?

### DIFF
--- a/crates/typlite/src/fixtures/integration/snaps/convert_tex@base.typ.snap
+++ b/crates/typlite/src/fixtures/integration/snaps/convert_tex@base.typ.snap
@@ -1,0 +1,23 @@
+---
+source: crates/typlite/src/tests.rs
+expression: "conv(world, ConvKind::LaTeX)"
+input_file: crates/typlite/src/fixtures/integration/base.typ
+---
+<!DOCTYPE html>
+<html>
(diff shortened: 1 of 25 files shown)


hongjr03:
> elements created by the user calling `html.frame` inside that have predefined display rules applied to them may lead to unexpected results, which needs some further discussions i think?
> 
> for example:
> 
> ```typst
> #html.frame(
>   grid(columns: 2, [hello], [world])
> )
> ```
> 
> will render nothing in result.

Has fixed by:
https://github.com/Myriad-Dreamin/tinymist/blob/f35d1056ffa909b7fe834e7faa8aa91888d49a00/crates/typlite/src/markdown.typ#L109-L115

PR #1734, opened 2025-05-02
feat: add change selections command

Myriad-Dreamin:
Adding a command to track user selections in the editor, then we can use the `focusing_selection` stored in the global state:

https://github.com/Myriad-Dreamin/tinymist/blob/8db172c6b40804a4cc368f6cb40b3392fba13997/crates/tinymist/src/server.rs#L79-L91

This is unfriendly for the editors that cannot make client-side extensions.

### DIFF
--- a/editors/vscode/src/extension.ts
+++ b/editors/vscode/src/extension.ts
@@ -152,11 +152,11 @@ async function languageActivate(context: IContext) {
   // Find first document to focus
   const editor = window.activeTextEditor;
   if (isTypstDocument(editor?.document)) {
-    focusDocPath(editor.document);
+    focusDoc(editor.document, editor);
   } else {
     window.visibleTextEditors.forEach((editor) => {
       if (isTypstDocument(editor.document)) {
-        focusDocPath(editor.document);
+        focusDoc(editor.document, editor);
       }
     });
   }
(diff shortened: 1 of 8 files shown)


editors/vscode/src/extension.ts
```diff
@@ -170,29 +170,16 @@ async function languageActivate(context: IContext) {
...
-        // console.log("not typst", langId, editor?.document.uri.fsPath);
-        return commandActivateDoc(undefined);
-      }
-      return commandActivateDoc(editor?.document);
+      return focusDoc(isTypstDocument(editor?.document) ? editor?.document : undefined, editor);
     }),
-  );
-  context.subscriptions.push(
     vscode.workspace.onDidOpenTextDocument((doc: vscode.TextDocument) => {
       if (doc.isUntitled && window.activeTextEditor?.document === doc) {
```
yassine-safraoui, line 176: I think the doc.isUntitled is a mistake, you probably meant !doc.isUntitled, because, just before on line 166 you return if the document is untitled which makes sense because you don't want to consider it.
```suggestion
      if (!doc.isUntitled && window.activeTextEditor?.document === doc) {
```

Myriad-Dreamin: Hello, I added comments to the code. I checked it and think of I'm not doing a bad refactor. Please check the comments:

https://github.com/Myriad-Dreamin/tinymist/blob/f4614eee70efd42b7d2fe133751d8daa5abc6122/editors/vscode/src/extension.ts#L208-L211

https://github.com/Myriad-Dreamin/tinymist/blob/f4614eee70efd42b7d2fe133751d8daa5abc6122/editors/vscode/src/extension.ts#L216-L223

PR #1737, opened 2025-05-03
feat: bump typstyle to v0.13.4 and add config for hard wrap

Enter-tainer:
TODO:

- [ ] should we use `enabled`, `disabled`, `auto` enum for hard wrapping?


Myriad-Dreamin:
we should remove the `auto`, as we didn't enable it for typstfmt as well. Just disable it by default.

crates/tinymist/src/config.rs
```diff
@@ -116,7 +116,8 @@ pub struct Config {
     pub formatter_print_width: Option<u32>,
     /// Sets the indent size (using space) for the formatter.
     pub formatter_indent_size: Option<u32>,
-
+    /// Sets the hard line wrapping mode for the formatter.
+    pub formatter_hard_wrap: Option<bool>,
```
Myriad-Dreamin, line 120: bikeshedding time, this should be "prose wrap". Existing editors and tools say "line wrap", "word wrap", or "prose wrap" and I haven't seen any configuration named "hard wrap". I understand "hard wrap" comes from "hard line wrap", but it becomes unclear when we omit the thing to wrap. Meanwhile, `formatterPrintWidth` already means hard line wrap that the formatter should fit in best effort, then "hard wrap" might be confusing. People may also ask: does "hard wrap" mean that whether "formatterPrintWidth" should be respect or not?

For the concrete name, I haven't decide a certain good choice:
- `formatterMarkupWrap`: wrap line in markup mode.
- `formatterMathWrap` or `formatterEquationWrap` may mean: wrap line in math mode or equations.
- `formatterCommentWrap` may mean: wrap line in comment mode.
- `formatterProseWrap`: wrap in prose which are in comment, math, or markup mode.
- `formatterCodeInProseWrap`: this is subtle that whether we will do it or not.
  ```
  // Show the following hash expression be affected by prose configuration?
  lorem #a-func-call-between-proses(a, b, c) ipsum.
  ```

Make things clear will benefit to extend the semantics in future. For example, prettier's prose wrap has mode more than `true` (always) or `false` (preserve?):

image

Enter-tainer: fixed, PTAL

PR #1797, opened 2025-06-05
chore: bump typstyle-core to v0.13.10 and update changelog

Enter-tainer:


Enter-tainer:
TODO: api has changed in v0.13.10

Issue #1845, opened 2025-06-23
typlite: Block elements are not rendered inside table (Typst → Markdown)


bluss:
Using typlite (v0.13.14)

Problem: Raw block is rendered before the table, not inside/part of the table.

Input document

````typ
#grid(
  columns: 2,
  [Header], [Row],
  ```
  Code line 1
  Code line 2
  ```,
  [Regular text]
)
````

Output document:

````markdown
```
Code line 1
Code line 2
```

| Header | Row |
| --- | --- |
|  | Regular text |
````

I think this problem applies to all non-inline elements.

Use case: using grid for placing things side by side in the layout.

bluss:
What I can see, the only option for embedding fenced code blocks is to write the table using html tags (`<table>`), not sure if this is something typlite can/should do automatically.

PR #1855, opened 2025-06-29
fix: table with mixed content (#1845)

hongjr03:


crates/typlite/src/fixtures/integration/snaps/convert@table.typ.snap
```diff
@@ -24,3 +25,19 @@ snapshot_kind: text
...
+<table>
+<thead>
+<tr>
+<th>Header</th>
+<th>Row</th>
+</tr>
+</thead>
+<tbody>
+<tr>
+<td><pre><code>Code line 1
```
Myriad-Dreamin: the raw language will be lost here.

hongjr03: because the corresponding test case does not include a language field. i added one in the latest commit and impl the html_writer_option field in writer_option.

Myriad-Dreamin: <table>
<thead>
<tr>
<th>Header</th>
<th>Row</th>
</tr>
</thead>
<tbody>
<tr>
<td>

```rs
fn test() {}
```

</td>
<td>Regular text</td>
</tr>
</tbody>
</table>

PR #1856, opened 2025-06-29
fix: typstyle text wrapping

andreaswilli:
Looks like the `formatterProseWrap` attribute was not actually read from the config so the default `false` was used.


Myriad-Dreamin:
@Enter-tainer @QuadnucYard Might have to confirm that the bug is fixed by installing the built extension and testing it locally.

PR #1879, opened 2025-07-06
Refactoring

adamnemecek:


Myriad-Dreamin:
We don't enfore `clippy::use_self`, for example `Ty::new` is better than `Self::new`

Myriad-Dreamin:
closed because this is likely to be a PR suspiciously made by LLM.

adamnemecek:
Why would it be an llm?

Myriad-Dreamin:
> Why would it be an llm?

Sorry if my judgment was mistaken, I recently receive some PRs which ran some automated tools (cargo in this PR).

PR #1903, opened 2025-07-11
feat(preview): add typst preview ng extension

Myriad-Dreamin:
- add an experimental Typst Preview NG provider extension
- keep the NG preview frontend source under tools/typst-preview-frontend-ng
- render preview data through worker-owned canvas frontend code
- use canvas cache keys in the default preview renderer
- update Tinymist preview dependencies to the typst.ts optimization branch


fiso64:
Any progress here? 
My device can't handle `onType` preview refreshing without input delays even when working on a small 3-page document, so I was considering implementing debounced preview refreshing (with a configurable delay), in case this PR is still a long way off from being merged.

Myriad-Dreamin:
> Any progress here? My device can't handle `onType` preview refreshing without input delays even when working on a small 3-page document, so I was considering implementing debounced preview refreshing (with a configurable delay), in case this PR is still a long way off from being merged.

it is slowly progressed. tho renderer was fully implemented we didn't merge it because we don't want to make regression. but recently we had idea of allowing customize preview's html. we may feel braver to publish another extension with bundles a renderer used in this pr. people using the extension could create issues to help development move faster.

Myriad-Dreamin:
@fiso64 since https://github.com/Myriad-Dreamin/tinymist/pull/2357, we will be able to use a different frontend. This will greatly help development:
- most users still use the default frontend.
- developers can make their owned implementations, like:
  - [OffscreenCanvas](https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvas) in this PR.
  - webgpu backend
  - QT backend
- we could switch the default implementation when we feel appropriate.

PR #1906, opened 2025-07-15
fix: update the docs link from `pattern` to `tiling`

YDX-2147483647:


Myriad-Dreamin:
This looks great. Does web have this issue?

YDX-2147483647:
> Does web have this issue?

Web means typst.app? I don't think there's a similar feature on typst.app.

Myriad-Dreamin:
> > Does web have this issue?
> 
> Web means typst.app? I don't think there's a similar feature on typst.app.

i thought ctrl+click on std function identifiers then webapp will bring you to the related page on the docs site.

YDX-2147483647:
> ctrl+click on std function

Oh, I didn't know… So typst.app does NOT have this issue.

YDX-2147483647:
I've just checked that `docs/src/lib.rs` skipped `pattern` explicitly.

https://github.com/typst/typst/blob/5661c20580584d86897d139cdeffd6fb1a9a6d5c/docs/src/lib.rs#L275-L278

```rust
    // Tiling would be duplicate otherwise.
    if category == Category::Visualize {
        skip.insert("pattern");
    }
```

These lines were added in https://github.com/typst/typst/pull/5812/files#diff-c1ae9d6224cd6fca410f72888401c94bdc6ffd1ffd12ac7629196abf74ed7617R275-R279, mixed in a large commit (+709 −448).

PR #1941, opened 2025-07-30
fix: correctly handle relative user-specified output paths in compile command

moeleak:
This change ensures that user-specified output paths in the compile command are properly handled, including relative paths.

Previously, the export task would reject relative output paths with an error. This change converts relative paths to absolute paths by joining them with the current working directory, allowing users to specify relative paths like 'output/test.pdf' when compiling documents.

The fix adds path normalization using PathClean and handles the conversion gracefully while maintaining the existing behavior for absolute paths.

Fixes the issue where commands like:
  tinymist compile resume-en.typ ./output/test.pdf
would fail with "output path is relative" error.

### DIFF
--- a/crates/tinymist/src/task/export.rs
+++ b/crates/tinymist/src/task/export.rs
@@ -10,6 +10,7 @@ use reflexo_typst::{Bytes, CompilationTask, ExportComputation};
 use tinymist_project::LspWorld;
 use tinymist_std::error::prelude::*;
 use tinymist_std::fs::paths::write_atomic;
+use tinymist_std::path::PathClean;
 use tinymist_std::typst::TypstDocument;
 use tinymist_task::{get_page_selection, ExportMarkdownTask, ExportTarget, PdfExport, TextExport};
 use tokio::sync::mpsc;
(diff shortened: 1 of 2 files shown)


crates/tinymist/src/task/export.rs
```diff
@@ -178,6 +179,12 @@ impl ExportTask {
         let Some(write_to) = output.substitute(&entry) else {
             return Ok(None);
         };
+        let write_to = if write_to.is_relative() {
+            let cwd = std::env::current_dir().context("failed to get current directory")?;
+            cwd.join(write_to).clean()
+        } else {
+            write_to.to_path_buf()
+        };
         if write_to.is_relative() {
```
Myriad-Dreamin, line 188: The correct "fix" should be that imo:
- Removing the conservative condition directly: `write_to.is_relative()`, which is not always true when we run the compile command.
- Adding tests to ensure that we still warn or error to the user via editors that `write_to` must be absolute. This could be added:
  - as unit tests in `crates/tinymist/src/config.rs`:
  - as e2e tests in either `editors/vscode/src/test/e2e/export.test.ts` or `editors/neovim/spec/export_spec.lua`.

Myriad-Dreamin:
Thanks for your contribution! I have setup e2e tests for you to validate fix and prevent regression, please check [tests/e2e/e2e/cli.rs](../blob/main/tests/e2e/e2e/cli.rs).

PR #1947, opened 2025-07-30
feat: compile tinymist-query targeting wasm32

Myriad-Dreamin:


Myriad-Dreamin:
@Momijiichigo Hi, have seen your work and it's awesome. I would like reuse some code from your branch if you authorize that.

PR #1987, opened 2025-08-04
fix: set VS Code default configuration for invertColors to "never"

BlueQuantumx:
Since #1807 , VS Code pops up a warning when starting tinymist: 
image

Because the default value was not set, it's trying to deserialize a `null`. I complete the corresponding configuration entry in the VS Code extension's package.json and package.other.json with default value "never".


Myriad-Dreamin:
besides fixing the default value in vscode, we need to handle the null case at the server side, because other clients like neovim may also pass a null value.

Myriad-Dreamin:
Thank you for your quick fix. I'll merge it tomorrow and publish a new release to fix it quickly.

PR #1997, opened 2025-08-05
docs: make translation of pinning clearer

rikhuijzer:
Thanks for making this awesome extension. I find it very useful.

This PR suggests a small rewording for a sentence. With this, I hope it becomes a bit clearer what the command does. This is also closer to Google Translate's translation of "将主文件固定到当前打开的文档".


Myriad-Dreamin:
Thank you!

Issue #1950, opened 2025-07-30
Add delay option for onType properties


migraine-user:
### Motivation

Setting preview to `onType` creates a lot of input lag. It would be nice if we can set a delay of some amount before the refresh is actually done. This would mitigate the input lag issue and also would prevent visual distractions while people are typing away.

### Description

If delay is set to 1000ms, when the extension detects that no typing has happened in the last 1000ms, refresh would be called. Otherwise, refresh would not be called.

PR #2057, opened 2025-08-21
fix(tinymist-project): invalidate cached snapshot after compile

jo3-l:
This PR fixes a latent cache invalidation issue with project compilation. I recommend reviewing commit-by-commit, even though this is a small change. Only the last commit changes behavior; the rest are cleaning up code to make it easier to see what is happening.

---

Here's what I think is the problem.

Project instance states (`ProjectInsState`) provide access to a snapshot of the latest compilation result. This snapshot is cached for performance:
```rust
match self.snapshot.as_ref() {
    Some(snap) if snap.world().revision() == self.verse.revision => snap.clone(),
    // otherwise recompute
}
```
Unfortunately, it turns out that the cached snapshot may be outdated even if the revision matches. (In practice, it will usually just lag one compilation behind, and I think this is why it hasn't caused any observable issues--one compilation behind is not that much, especially if the user is continuously typing.)

To see the issue, consider the following sequence of steps. Assume initially all revisions are `0`.
1. User triggers a recompilation, say by typing into the file.
     - This triggers an `Interrupt::Memory` or similar.
2. The changes are applied to the vfs and recompilation begins.
     - `verse.revision` is incremented to `verse.revision = 1`.
3. A new snapshot is requested.
    - The project compiler checks the cached snapshot first, which still has `snap.world().revision() == 0`.
    - Since `verse.revision = 1` was already incremented before the compilation finished, a new snapshot is created and cached.
    - The cached snapshot now has `snap.world().revision() == 1`, even though nothing changed.
4. Compilation finishes, and some data from the resulting compilation artifact is recorded.
5. Another snapshot is requested.
    - Since the snapshot was recomputed already in (3), the revisions appears to match and the outdated, cached snapshot is returned.

The fix here is to just clear the cached snapshot in (4) after recompilation finishes.

### DIFF
--- a/crates/tinymist-project/src/compiler.rs
+++ b/crates/tinymist-project/src/compiler.rs
@@ -405,11 +405,11 @@ impl<F: CompilerFeat + Send + Sync + 'static, Ext: Default + 'static> ProjectCom
             ext: Default::default(),
             verse,
             reason: no_reason(),
-            snapshot: None,
             handler,
             export_target,
             deps: Default::default(),
             latest_compilation: None,
+            cached_snapshot: None,
         }
     }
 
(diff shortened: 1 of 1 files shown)


Myriad-Dreamin:
I'll mock some interface and add unit tests for this important component. This will help test some other features in plan for example:
- #998 
- #1950

PR #2062, opened 2025-08-21
feat: offer quickfix to add spaces separating letters in unknown math var

jo3-l:
Examples:

| Input | Suggested fix |
| -- | -- |
| `$xyz$` | `$x y z$` |
| `$a_ij$` | `$a_(i j)$` |
| `$a_(ij)$` | `$a_(i j)$` |
| `$xy/z$` | `$(x y)/z$` |

If the unknown identifier appears as a subscript or as the numerator/denominator of a fraction, we parenthesize the suggested fix. For example, `a_ij` turns into `a_(i j)`, not `a_i j`, because the latter is unlikely to be what the user intended.


crates/tinymist-query/src/code_action.rs
```diff
@@ -116,4 +120,32 @@ mod tests {
...
+        // Run the linter for additional diagnostics as well.
+        let diagnostics = DiagWorker::new(ctx)
+            .check()
+            .convert_all(compiler_diagnostics)
+            .into_values()
+            .flatten();
+
+        CodeActionContext {
+            diagnostics: diagnostics
+                .filter(|diag| ranges_overlap(&diag.range, request_range))
```
jo3-l, line 146: The filtering here matches the LSP spec (and VS Code behavior):

> `diagnostics`: An array of diagnostics known on the client side **overlapping the range provided** to the `textDocument/codeAction` request [...]

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeActionContext

crates/tinymist-query/src/code_action.rs
```diff
@@ -116,4 +120,32 @@ mod tests {
...
+                .collect(),
+            only: None,
+            trigger_kind: None,
+        }
+    }
+
+    fn ranges_overlap(r1: &LspRange, r2: &LspRange) -> bool {
```
jo3-l, line 153: If there's already an existing implementation of range overlap elsewhere in the project, let me know and I'll switch to it. (I couldn't find anything when I searched, so I just rolled my own.)

crates/tinymist-query/src/code_action.rs
```diff
@@ -116,4 +120,33 @@ mod tests {
...
+        let Warned { output, warnings } = typst::compile::<PagedDocument>(&ctx.world);
+        let errors = output.err().unwrap_or_default();
+        let compiler_diagnostics = warnings.iter().chain(errors.iter());
+
+        // Run the linter for additional diagnostics as well.
+        let diagnostics = DiagWorker::new(ctx)
+            .check(source)
+            .convert_all(compiler_diagnostics)
+            .into_values()
+            .flatten();
```
jo3-l, line 138: Question: This code is more finnicky than I would like, because `DiagWorker` is designed to handle diagnostics across multiple files, whereas we only care about diagnostics in a specific file here. (The parts I don't like: I had to separate `check(source)` and `full_check()`, and we have to do this `.into_values().flatten()` thing at the end.)

Do you think we should refactor the API of `DiagWorker` to make it easier to compute diagnostics for just one file, or is it fine to leave as is?

crates/tinymist-query/src/code_action.rs
Myriad-Dreamin: we can add comments into code If we think of some references are useful for people.

crates/tinymist-query/src/code_action.rs
Myriad-Dreamin: We can reference your comment here from another PR, and then discuss them there. Let's left `DiagWorker` things as is in this PR.

crates/tinymist-query/src/code_action.rs
Myriad-Dreamin: I don't remeber whether we have similar function, but this impl seems incorrect.

example: `r1 = [1, 2), r2 = [2, 3)`, and 
```rs
ranges_overlap([1, 2), [2, 3))
  !(2 < 2 || 3 < 1)
  !(false || false)
  !false
  true
```

I found it using the regex to grep the codebase `end.*start` to get previous similar code:

https://github.com/Myriad-Dreamin/tinymist/blob/05280aec4d23d144620fd0c2596251483c9c7f0c/crates/tinymist-query/src/inlay_hint.rs#L92-L94

crates/tinymist-query/src/code_action.rs
jo3-l: Good suggestion, thanks. Added a comment in https://github.com/Myriad-Dreamin/tinymist/pull/2062/commits/da23816a79cb945d62876fbbff7b2c0d5a0ff422.

crates/tinymist-query/src/code_action.rs
jo3-l: Hah, knew I should've checked this a little more carefully. As you note, the comparisons should be inclusive. (Hopefully) fixed in https://github.com/Myriad-Dreamin/tinymist/pull/2062/commits/da23816a79cb945d62876fbbff7b2c0d5a0ff422.

crates/tinymist-query/src/analysis/code_action.rs
```diff
@@ -198,6 +207,34 @@ impl<'a> CodeActionWorker<'a> {
         Some(())
     }
 
+    /// Add spaces between letters in an unknown math identifier: `$xyz$` -> `$x y z$`.
```
Myriad-Dreamin, line 210: The such comment sentences should be in indicate mood. For example,

```patch
- Add spaces between letters in an unknown math identifier: `$xyz$` -> `$x y z$`.
+ Adds spaces between letters in an unknown math identifier: `$xyz$` -> `$x y z$`.
```

jo3-l: Got it, will edit my other PRs accordingly. Sorry, I've been trying to use indicative in line with the rest of the codebase, but I'm too used to imperative and habits die hard :-)

jo3-l:
Hm. Sorry about the lint failures here, they slipped my attention locally since they show up as warnings, not errors, in my editor. I'm happy to put a PR fixing them later unless you get to it first.

Not too sure how to deal with the unused `DiagWorker::check` method, since it _is_ used, just in a test-only module. We can certainly mark it as `#[cfg(test)]` but that feels a little nasty.

PR #2065, opened 2025-08-22
feat: new lint warning for unknown math vars

jo3-l:
**Stacked on top of #2062; please review that PR first.**

Add a lint warning for unknown math variables[^1], powered by Tinymist's existing reference analysis. This allows users to see all such undefined variables at once if they have the linter enabled, instead of just the first error from the compiler. The autofix from #2062 is also applicable to these warnings.

Implementation notes:
- The generated diagnostic tries to closely match the corresponding Typst compiler error, with one deliberate change: to differentiate the Tinymist warning and the compiler error, we use the message `potentially unknown variable: ...` instead of `unknown variable: ...`. I am not the biggest fan of this choice, but I think it is very important that users don't blame the Typst compiler for warnings that we generate; changing the message so that it isn't an exact clone is the best way I thought of for now.
- We avoid duplicating a warning if the compiler has already generated an error for a given identifier by computing a set of `KnownLintIssues` from the compiler diagnostics and threading this information through the lint routines.

[^1]: If users like this warning, we can extend it later to apply to undefined variables outside of math as well.


crates/tinymist-lint/src/lib.rs
```diff
@@ -31,28 +34,74 @@ pub struct LintInfo {
...
+        revision: ei.revision,
+        fid: ei.fid,
         diagnostics,
     }
 }
 
-struct Linter<'w> {
+/// Information about issues the linter checks for that will already be reported to the user via
+/// other means (such as compiler diagnostics), to avoid duplicating warnings.
+pub struct KnownLintIssues {
```
jo3-l: We could consider computing `KnownLintIssues` lazily (i.e., only compute `unknown_vars` from the compiler diagnostics when we check for a duplicate warning.) It's kind of annoying though because that forces us to store the compiler diagnostics in `KnownLintIssues`, which requires either an allocation from collecting into a Vec or a generic type parameter (since the iterator type is not concrete.)

IMO it's not worth the complexity since usually there are not many compiler diagnostics, so the current code computes eagerly in the `from_compiler_diagnostics` constructor.

crates/tinymist-lint/src/lib.rs
```diff
@@ -386,18 +435,64 @@ impl DataFlowVisitor for Linter<'_> {
...
             self.check_variable_font(expr.args().items());
         }
+        self.exprs(expr.args().to_untyped().exprs().chain(expr.callee().once()));
+        Some(())
+    }
+
+    fn math_ident(&mut self, ident: ast::MathIdent<'_>) -> Option<()> {
+        let resolved = self.ei.get_def(&Interned::new(Decl::math_ident_ref(ident)));
+        let is_defined = match resolved {
+            Some(Expr::Ref(refs)) => refs.root.is_some() || refs.term.is_some(),
```
jo3-l: Not sure if this is the right check. Empirically, it seems to do the right thing on a couple of larger documents I tested. However, I fully admit I don't understand the meaning of the various fields in `RefExpr` at the moment, so I ended up with this condition purely from trial and error.

jo3-l:
This is now ready for review. ~~I expect Clippy is going to fail; I will fix the warnings once #2096/a similar fix is merged.~~ Fixed.

crates/tinymist-lint/src/lib.rs
Myriad-Dreamin: Does the comment recently generated by LLM help understand the code? I look it again and find it a bit lengthy, but I can continue improve it.

https://github.com/Myriad-Dreamin/tinymist/blob/532f25abe97cc646821e82c1a339c87c3c664f38/crates/tinymist-analysis/src/syntax/def.rs#L1015-L1030

Myriad-Dreamin:
Haven't started review yet. I add a short comment that may address some existing problem about this PR.

> Implementation notes:
> 
> * The generated diagnostic tries to closely match the corresponding Typst compiler error, with one deliberate change: to differentiate the Tinymist warning and the compiler error, we use the message `potentially unknown variable: ...` instead of `unknown variable: ...`. I am not the biggest fan of this choice, but I think it is very important that users don't blame the Typst compiler for warnings that we generate; changing the message so that it isn't an exact clone is the best way I thought of for now.
> * We avoid duplicating a warning if the compiler has already generated an error for a given identifier by computing a set of `KnownLintIssues` from the compiler diagnostics and threading this information through the lint routines.
> 
> ## Footnotes
> 1. If users like this warning, we can extend it later to apply to undefined variables outside of math as well. [↩](#user-content-fnref-1-072eb8a41c7791f4e64dd3c5ac1768bf)

We can also change the source from `typst` to `tinymist` (or `tinymist-lint` to help users get that it is a linting diagnostic message).

https://github.com/Myriad-Dreamin/tinymist/blob/6e386e2838ffa1f36e71604383279be1bf1b8b55/crates/tinymist-query/src/diagnostics.rs#L124

crates/tinymist-lint/src/lib.rs
jo3-l: I read those comments when I was originally writing the code, but I'm mainly interested in when `type` and `root` are set (or, conversely, when they are `None`.) The comments don't really go into that detail AFAICS?

Myriad-Dreamin:
I added some commits, the title of commits explains the fixed minor issue. One main issue is `known_issues` was not added to hasher to determine whether to cache `compute` or not.

PR #2069, opened 2025-08-23
build: make use of async io of js to speed up vscode builds

Myriad-Dreamin:


scripts/build-l10n.mjs
```diff
@@ -53,36 +56,38 @@ function translate(output, kind, inputs) {
...
   const langDir = path.resolve(projectRoot, output);
-  fs.mkdirSync(langDir, { recursive: true });
+  await fs.mkdir(langDir, { recursive: true });
 
   const langEnPath = `${langDir}/${kind}.json`;
   const langEnData = translations["en"];
-  fs.writeFileSync(langEnPath, JSON.stringify(langEnData, null, 2));
+  await fs.writeFile(langEnPath, JSON.stringify(langEnData, null, 2));
 
   for (let lang of langRest) {
```
jo3-l, line 65: Driveby comment: I think the IO in this loop is still running sequentially, but it looks like it could be parallelized via `Promise.all`. That might give a further speedup, depending on how many languages there are.

Myriad-Dreamin: running async io is to provide schedule points to be preempted by other async tasks (the other tasks emitted by `Promise.all`. The js scripting is always sequential probably.

jo3-l: I meant that the promise returned from `writeFile` resolves only when the write completes, so if you await them in sequence, the second file write will only start after the first completes and so on. OTOH if you `Promise.all` the file write promises, they can all start at once. I don't think this optimization will end up changing anything at this scale, though, so it was just a driveby suggestion :-)

Myriad-Dreamin: Oh, you may mean the writeFile in the for loop instead of that before the loop. GitHub only showed the L60-65 and confused me. Yeah you should be correct. From stats of my windows PC, the time used before PR and after PR are 117 seconds and 17.5 seconds. Continuing spotting the rest 17.5 seconds, the critical bottleneck `cargo build -p tinymist-cli` took 17.05 seconds among tasks. This means that we must optimize cargo debug build then to continue speed up time to launch extension debugging. As for the for loop, that doesn't matter a lot.

jo3-l: Makes sense (GitHub UI is not the greatest sometimes), and agree that the cargo build is definitely the biggest bottleneck. I assume you are already aware of this, but if you want to speed up the debug build, it might be useful to see if switching to https://github.com/rust-lang/rustc_codegen_cranelift is viable. LLVM can be very slow, and I've seen [reports that Cranelift is noticeably faster](https://old.reddit.com/r/rust/comments/1bgyo8a/try_cranelift_codegen_backend_for_faster_compile/) on some projects. The downside is one can only use this on a nightly toolchain.

FWIW I switched my local dev build to use the Cranelift backend because I'm not on a super beefy laptop, and subjectively it feels snappier. But I haven't actually compared before/after carefully.

PR #2094, opened 2025-08-29
refactor: preview page remove extra space and do not invert background

hyrious:
| Before | After |
| ------ | ----- |
| image | image |

1. The `svg` element by default has `display: inline`, which adds extra space at bottom (line height), change that to `block` can remove the space.
2. The `svg` element and its outer `div` both set background color, which will be inverted to another color that are not equal to the whole page's background (I was setting `--main-margin` to see the outside). This PR prevents the background from getting inverted.

There's another tiny thing: `#typst-container.mode-doc` has `margin-top: -1px` which shifts the whole Typst doc 1px up. That will cause the page always has 1px margin at bottom because the SVG element has the full height. One possible fix is to always `- 1px` on the SVG `height` attribute but I'm not sure if this affects some calculation. So I'm not adding this trivial change.


Myriad-Dreamin (approved):
I confirm that it works. Thank you!

PR #2095, opened 2025-08-29
feat: hover tooltip for package import

QuadnucYard:
This PR adds package information as the definition of module import in the package import. It also shows available versions.

image

Its replace-version function imitates Dependi. We may need to add trailing decorations for a quick upgrade, or add code actions to upgrade hover/all import versions of the given package.

### DIFF
--- a/crates/tinymist-query/src/fixtures/hover/snaps/test@package_import_no_upgrade.typ.snap
+++ b/crates/tinymist-query/src/fixtures/hover/snaps/test@package_import_no_upgrade.typ.snap
@@ -5,10 +5,10 @@ input_file: crates/tinymist-query/src/fixtures/hover/package_import_no_upgrade.t
 ---
 Range: 0:29:0:53
 
-🌌 [Universe](https://typst.app/universe/package/example)
+[Universe](https://typst.app/universe/package/example)
 
 **Package:** `@preview/example:0.1.1`
-✅ **Up to date** (latest version)
+**Up to date** (latest version)
 
 **Description:** example package (mock).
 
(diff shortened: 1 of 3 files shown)


crates/tinymist-query/src/hover.rs
```diff
@@ -214,6 +217,173 @@ impl HoverWorker<'_> {
...
+            // Extract and sort versions
+            all_versions = package_entries
+                .iter()
+                .map(|(spec, _)| spec.version)
+                .sorted()
+                .rev()
+                .collect(); // Show newest first
+
+            latest_version = registry
+                .determine_latest_version(&package_spec.versionless())
```
Myriad-Dreamin: This requires network access and will cause nixpkgs test to fail. We may either skip tests (mark as todo) or add some interfaces that don't access network during testing.

crates/tinymist-query/src/package.rs
```diff
@@ -173,7 +210,22 @@ pub fn list_package_by_namespace(
                     name: package.file_name().to_string_lossy().into(),
                     version,
                 };
-                packages.push((package_version_path, spec));
+                let manifest_id = typst::syntax::FileId::new(
+                    Some(spec.clone()),
+                    typst::syntax::VirtualPath::new("typst.toml"),
+                );
+                let Some(manifest) =
+                    once_log(get_manifest(world, manifest_id), "read package manifest")
```
Myriad-Dreamin, line 176: We must reduce cost to read manifests. This requires heavy IO (may read hundreds of packages), and `package_completions` invokes it without caching.

Possible improvement paths:
1. cache aggressively: since we know packages are rarely changed, we can read package manifest once and never clean up the cache. This must be done in this PR.
2. watch manifest file changes: left in future PR.

PR #2102, opened 2025-09-03
feat: add js package registry support for tinymist-wasm

ParaN3xus:


crates/tinymist/src/project.rs
```diff
@@ -291,6 +320,7 @@ impl ProjectInsStateExt {
 
         let pending_reasons = self.pending_reasons.exclude(self.emitted_reasons);
         if !pending_reasons.any() {
+            log::info!("skipping emit bcuz there's no reason");
```
Myriad-Dreamin: Please remove the logs that may flood the output channel when running in system.

crates/tinymist/src/project.rs
```diff
@@ -280,6 +308,7 @@ impl ProjectInsStateExt {
         handler: &dyn CompileHandler<LspCompilerFeat, ProjectInsStateExt>,
     ) -> bool {
         let Some(last_compilation) = self.last_compilation.as_ref() else {
+            log::info!("skipping emit bcuz there's no last_compilation");
```
Myriad-Dreamin: Please remove the logs that may flood the output channel when running in system.

crates/tinymist/src/project.rs
```diff
@@ -17,6 +17,8 @@
 //!
 //! The [`CompileHandlerImpl`] will push information to other actors.
 
+#![allow(missing_docs)]
```
Myriad-Dreamin: remove the macro directive.

crates/tinymist-project/Cargo.toml
```diff
@@ -35,6 +35,8 @@ toml = { workspace = true, optional = true }
 typst.workspace = true
 typst-assets.workspace = true
 notify.workspace = true
+js-sys = { workspace = true }
```
Myriad-Dreamin: this should be an optional dependency which enabled only if `feature = web`. There are similar code in other cargo.toml of the repository.

PR #2111, opened 2025-09-10
feat: scroll preview with hjkl

### LINKED ISSUE #2107: Scrolll preview with vim keybinds
### Motivation

When using neovim or helix with some kind of tiling window manager setup, it is inconvenient to change hand position or reach for the mouse to scroll the preview. It would be nice to use the home row keys to navigate the preview.

### Description

My idea is for tinymist to handle the `h`, `j`, `k` and `l` keys to do scroll up/down for j/k and next/previous page for h/l. I would hope this is relatively straightforward to implement using javascript in the page the preview serves.

The problem of changing hand positions would be somewhat mitigated by making `scrollSync` work outside of vscode, but I don't know if it's possible/feasible and, regardless, having this in tynimist seems like a useful thing to have.

### More Examples/Questions

One question is whether to add an option to disable it. I think it makes sense to have it enabled by default, I don't see how it would do much harm and it would be nice for people using their muscle memory. 

Another option that could be useful would be to specify how much to scroll by with each key, but maybe that's already too much and I assume we don't want to bloat unnecessarily the configuration options. 

I don't think there is any need to add any other vim keybinds other than hjkl.

odilf:
Closes #2107 

Sorry for not getting confirmation on the issue first, but I had some free time and thought it would be easy to implement and, indeed, it is just a few lines of code. I'll attach a video of how it looks. 

https://github.com/user-attachments/assets/60e6f7ef-cdff-4f85-a1e8-385058f87966

For the small nudges I think the instant jump is better, even though it looks kind of janky. The problem with the smooth scroll is that you can't press it repeatedly and I think that you use `jk` for small adjustments where it's clear what is being scrolled. However, for the big hl jumps I think the smooth scroll is better to not disorient you. 

Happy to change anything, of course :)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -173,9 +173,9 @@ checksum = "7c02d123df017efcdfbd739ef81735b36c5ba83ec3c59c80a9d7ecc718f92e50"
 
 [[package]]
 name = "async-trait"
-version = "0.1.88"
+version = "0.1.89"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e539d3fca749fcee5236ab05e93a52867dd549cc157c8cb7f99595f3cedffdb5"
+checksum = "9035ad2d096bed7955a320ee7e2230574d28fd3c3a0f186cbea1ff3c7eed5dbb"
 dependencies = [
  "proc-macro2",
  "quote",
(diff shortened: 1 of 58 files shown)


Myriad-Dreamin:
Could we add description of these keys to `typst-help-panel` in the `index.html`? I see `h` was used to show help panel. Since they are rarely used, I'm okay to change existing key bindings.

Myriad-Dreamin:
Having vim key bindings to control preview does make us feel like a real hacker. Looks very nice.

odilf:
Sorry for the delay!

> Could we add description of these keys to `typst-help-panel` in the `index.html`? I see `h` was used to show help panel. Since they are rarely used, I'm okay to change existing key bindings.

Hmm, is this supposed to work on lsp preview? I'm using helix, but pressing `h` on the preview doesn't seem to do anything. The code does seem easy to change so I'll do the change, it's only that I won't be able to test it.

Also, I assume the key to show the help panel needs to change to some other key, does `?` sound good?

Myriad-Dreamin:
We will go back in tinymist v0.14, because tinymist v0.13 only accepts bug fixes now.

Myriad-Dreamin:
> Sorry for the delay!
> 
> > Could we add description of these keys to `typst-help-panel` in the `index.html`? I see `h` was used to show help panel. Since they are rarely used, I'm okay to change existing key bindings.
> 
> Hmm, is this supposed to work on lsp preview? I'm using helix, but pressing `h` on the preview doesn't seem to do anything. The code does seem easy to change so I'll do the change, it's only that I won't be able to test it.
> 
> Also, I assume the key to show the help panel needs to change to some other key, does `?` sound good?

The changes look good to me, but I would like to test it in local by myself and merge it then.

Myriad-Dreamin (approved):

Issue #2021, opened 2025-08-10
[VSCode] Symbols list incomplete


e-zz:
(Please refer to #2019 for my system specifications. Tested also on a linux machine.)

Hi 
The symbols list panel in vscode (invoked by `ctrl+shift+o` or `ctrl+p+@`) is very often incomplete.
For example, a file `test.typ` has only one variable
```typst

#let c_bg = rgb("#fafafa")
```
We can see our only variable in the file `c_bg` 

<img width="250" height="115" alt="Image" src="https://github.com/user-attachments/assets/6bafe6df-d0f2-4d66-8561-5f28e19a12e5" />

But with a commented line above the variable, like
```typst
//
#let c_bg = rgb("#fafafa")
```
The variable `c_bg` will be invisible in the symbol list 

<img width="250" height="115" alt="Image" src="https://github.com/user-attachments/assets/03a28492-29ed-473a-a15f-3d1e56098934" />

PR #2116, opened 2025-09-12
fix: correct the lexical hierarchy of comment group

BlueQuantumx:
Hierarchy of comment group is always wrong previously, causing #2021.

This PR fixed it and added a test case.

BTW, the changes in the lexical hierarchy module influenced a folding range test case, but it's just changed the order of items, which doesn't affect the correctness (the order is even more correct).


Myriad-Dreamin (approved):

Myriad-Dreamin:
Nice.

PR #2166, opened 2025-10-10
fix: re-correct order of scope merging & make test stronger

BlueQuantumx:
Sadly the #2154 is still false. This PR fixed it and made test stronger.


Myriad-Dreamin:
We could add more tests when we make some mistake repeatedly. It doesn't prevent merge as CI is passing.

PR #2171, opened 2025-10-11
chore: switch cmark-writer dependency to local path

hongjr03:
Further fix and refactor may base on this pr.

also refactor: centralize CommonMark spacing and fix inline custom node handling

- add `FormatPolicy` to manage trailing newline / blank-line rules and expose prepare_block_sequence
- route document, blockquote, and list content through the policy so nested blocks get consistent spacing
- adjust `Node::is_block` / `is_inline` to respect `CustomNode::is_block`, preventing inline custom nodes from forcing HTML fallbacks
- ensure table rendering stays in Markdown by recognizing inline custom cells correctly

### DIFF
--- a/crates/cmark-writer-macros/LICENSE
+++ b/crates/cmark-writer-macros/LICENSE
@@ -1,0 +1,9 @@
+MIT License
+
+Copyright 2025 Hong Jiarong
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the “Software”), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+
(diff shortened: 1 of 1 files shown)


crates/cmark-writer/README.md
```diff
@@ -0,0 +1,172 @@
+# cmark-writer
+
+[![CI Status](https://github.com/hongjr03/cmark-writer/workflows/CI/badge.svg)](https://github.com/hongjr03/cmark-writer/actions)
+[![Crates.io](https://img.shields.io/crates/v/cmark-writer.svg)](https://crates.io/crates/cmark-writer)
+[![License: MIT](https://img.shields.io/badge/License-MIT-green.svg)](LICENSE)
```
Myriad-Dreamin: Add your license for all source files in the cmark-writer crate.

crates/cmark-writer/src/writer/html/core/mod.rs
```diff
@@ -0,0 +1,497 @@
...
+    /// ensure escaping.
+    pub fn write_trusted_html(&mut self, html: &str) -> HtmlWriteResult<()> {
+        self.ensure_tag_closed()?;
+        self.buffer.push_str(html);
+        Ok(())
+    }
+
+    /// Writes HTML content that may contain characters requiring escaping.
+    ///
+    /// This is a semantic alias for [`text`], making call sites explicit about
```
Myriad-Dreamin: fix warnings like this.

crates/cmark-writer-macros/Cargo.toml
Myriad-Dreamin, line 1: move this crate to `/crates/` folder.

crates/cmark-writer/README.md
Myriad-Dreamin: Is there any missing? I guess this is reported because no all markdown viewer will handle a file path link. We should change it to a HTTP url I guess.

Issue #1844, opened 2025-06-23
typlite: HTML rendering not implemented for Custom Node: typlite::common::VerbatimNode (When nested)


bluss:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Platform

x64 Linux (linux-x64, Most Common)

### Editor

Neovim

### Editor Version

N/A

### Describe the bug

Using typlite to convert Typst to Markdown.

Inside a show rule, I could not use `m1verbatim` to emit verbatim Markdown syntax.

<details>
<summary>bugrep.typ</summary>

`````typ
#let x-target = sys.inputs.at("x-target", default: "pdf")
#let show-if(cond, func) = body => if cond { func(body) } else { body }
#show grid: show-if(x-target == "md", it => {
  let columns = it.columns.len()
  html.elem("table", attrs: (x: "test"), {
    for row in it.children.chunks(columns) {
      html.elem("tr", {
        row.map(cell => html.elem("td", cell)).join()
      })
    }
  })
})

#show raw.where(block: true): show-if(x-target == "md", it => {
  // verbatim output: fail when nested
  html.elem("m1verbatim", attrs: (src: "````\n" + it.text + "\n````"))
})

= Reproducer document

== Example

#let src = {
````typ

== Under the Greenwood Tree
by Shakespeare.

#some-function(```
Under the greenwood tree #emoji.tree
Who loves to lie with me,
...
```)
````
}

Just the source (ok)

#src

Source in grid (fail)
// HTML rendering not implemented for Custom Node: typlite::common::VerbatimNode

#grid(
  columns: 2,
  [Placeholder for \ separately rendered SVG],
  src,
)
`````

</details>

Source, generated Markdown, Generated HTML are here: https://gist.github.com/bluss/b06cb65a93a741555670868b104c892c

### Server Logs

```log
Paste your logs here
```

### Browser Logs

```log
Paste your logs here
```

PR #2172, opened 2025-10-11
fix: enable HTML rendering for VerbatimNode and add regression test

hongjr03:
fix https://github.com/Myriad-Dreamin/tinymist/issues/1844


bluss:
nice, thanks!

Issue #1852, opened 2025-06-27
typlite: Links are not working


mindreframer:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Platform

Apple Silicon macOS (darwin-arm64, Most Common)

### Editor

VS Cod(e,ium)

### Editor Version

Example document: 

```
#quote(block: true)[
  #strong[Author name]
]

MY BOOK

#quote(block: true)[
  #strong[A quote]
]

= Content
<содержание>
Глава 1


#link(<first>)[FIRST]

#quote(block: true)[
  Глава 1
]

= FIRST
<first>
#quote(block: true)[
  #emph[Some text]
]

```

### Describe the bug

```bash
error: expected string, found label
   ┌─ @local/markdown:0.1.0/lib.typ:41:11
   │
41 │     attrs: (dest: dest),
   │            ^^^^^^^^^^^^

help: error occurred in this call of function `md-link`
    ┌─ @local/markdown:0.1.0/lib.typ:194:36
    │
194 │   show link: it => if-not-paged(it, md-link(dest: it.dest, it.body))
    │                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

help: error occurred while applying show rule to this link
   ┌─ /Users/roman/Desktop/WordToHTML/docs/book2/__wrap_md_main.typ:18:1
   │
18 │ #link(<first>)[FIRST]
   │  ^^^^^^^^^^^^^^^^^^^^

```

### Server Logs

```log
Paste your logs here
```

### Browser Logs

```log
Paste your logs here
```

### Additional context

typlite version: v0.13.14 (https://github.com/Myriad-Dreamin/tinymist/releases/tag/v0.13.14)

Compilation with typst works fine.

PR #2176, opened 2025-10-11
fix: normalize typst link destinations in typlite

hongjr03:
fix https://github.com/Myriad-Dreamin/tinymist/issues/1852

however we don't process label in `markdown.typ`, so that it's not guaranteed that every dest is able to jump.


Myriad-Dreamin:
We may only process the case where headings are the destinations?

hongjr03:
Since typst 0.14 [supports intra-doc links in HTML](https://github.com/typst/typst/pull/6602), maybe we should not fix the issue until the 0.14 library merged into the main branch.

PR #2182, opened 2025-10-14
feat: export tool with page/text preview features

QuadnucYard:
Added an editor tool to export and preview documents, with all configurable options.

Not supported: output path customization (requires changes in server)

image
image

### DIFF
--- a/tests/e2e/e2e/lsp.rs
+++ b/tests/e2e/e2e/lsp.rs
@@ -44,7 +44,7 @@ fn test_lsp() {
         });
 
         let hash = replay_log(&root.join("vscode-syntax-only"));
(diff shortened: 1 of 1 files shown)


Myriad-Dreamin:
There are following points to enhance, but we don't have to achieve them in this PR.
1. more PDF options: pdf standard, no pdf tags.
1. Add a top-level codelens to open the export tool
1. Move export button before the thumbnails
1. change name "generate preview" to "flush preview". It must be put together with the export button so that people can easily find the flush button.
1. Option to set column of thumbnails. When column is set to 2, it is suitable to check whether PDF output is nice typeset as double-sided document.  
1. diagnostics are not styled, but we could instead navigate people to the problem panel if there are any diagnostics.
1. Hint that the document must pass compile with [HTML export](https://typst.app/docs/reference/html/), when errors: "document set rules are not allowed inside of containers"

PR #2190, opened 2025-10-24
Fix URL to table guide

Rheddes:


ParaN3xus:
This is a test snap that will be updated once tinymist releases its first stable version supporting typst 0.14.0.

Myriad-Dreamin:
It has been updated by #2201.

PR #2278, opened 2025-11-24
feat: customize paste snippet in vscode

Myriad-Dreamin:
Add a field `edits` to control the snippet used for generating pasted text.

Example (default behavior when pasting image):

```typ
(
  edits: (
    kind: "by-mode",
    math: `#image("image.png")`.text,
    markup: `#image("image.png")`.text,
    code: `image("image.png")`.text,
    string: "image.png",
    comment: "",
    raw: "",
  )
)
```


Rickyneer:
Could it, just like with #2238 have custom scripting to create your own "resolver"? I.e. based on custom behavior, we can have it become `#bytes` instead of `#image`

PR #2320, opened 2025-12-20
feat: handle `onEnter` only when server is active and add timeout

QuadnucYard:
Previously, before tinymist extension is fully activated, enter key events are blocked and handled together with significant latency.

Now we ensure the keybinding only becomes active after the command is registered, and skip the request when the server has not been started. Besides, we added a cancellation timeout to avoid stalling.

### DIFF
--- a/editors/vscode/src/extension.shared.ts
+++ b/editors/vscode/src/extension.shared.ts
@@ -130,24 +130,12 @@ export async function tinymistActivate(
   configureEditorAndLanguage(context, trait);
 
   // Initializes language client
-  if (extensionState.features.lsp) {
-    const executable = tinymist.probeEnvPath("tinymist.serverPath", config.serverPath);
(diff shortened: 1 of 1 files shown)


Myriad-Dreamin:
can we also detect if the server is healthy and raise warning message in extension scripts? We didn't fix the "onEnter" error (I'm lazy) because it indicates the server is gone for some reason.

Myriad-Dreamin (approved):

editors/vscode/src/lsp.ts
```diff
@@ -123,12 +124,44 @@ export class LanguageState {
       await this.client.stop();
       this.client = undefined;
     }
+    // Reset server readiness flag so other code doesn't assume a running server
+    try {
```
Myriad-Dreamin: This try-catch is confusing.

PR #2332, opened 2025-12-30
feat: edit compiler inputs and reset options in the exporter

QuadnucYard:


crates/tinymist/src/cmd/export.rs
```diff
@@ -230,9 +236,11 @@ impl ServerState {
     pub fn export(&mut self, task: ProjectTask, mut args: Vec<JsonValue>) -> ScheduleResult {
         let path = get_arg!(args[0] as PathBuf);
         let action_opts = get_arg_or_default!(args[2] as ExportActionOpts);
+        let common_opts = get_arg_or_default!(args[3] as ExportCommonOpts);
         let write = action_opts.write.unwrap_or(true);
         let open = action_opts.open;
+        let inputs = common_opts.inputs;
```
Myriad-Dreamin, line 242: we should get `inputs` from `args[1]` (export options) instead of a new `args[3]`.

crates/tinymist/src/task/export.rs
```diff
@@ -69,9 +71,23 @@ impl ServerState {
 
         let snap = self.snapshot().map_err(internal_error)?;
         just_future(async move {
+            let inputs = inputs.map(|inputs| {
+                // merge with existing inputs
```
Myriad-Dreamin, line 75: If we merge with existing inputs, how could a user remove some existing `sys.inputs` by configuration?

PR #2360, opened 2026-01-20
feat: allow previewing a specific Typst file without opening it in an editor tab

sticks-stuff:
This is primarily designed to make it easier for other extensions to preview a specific `.typ` file without first having to focus the file, then running `typst-preview.previewFile`.

However, it still works and provides a quick picker of `typ` files if manually invoked by the user through the command palette.

image

As far as I can tell, this has no regressions in terms of functionality, though the `preview-compat` stuff and `isCompat` stuff was not tested extensively.

### DIFF
--- a/editors/vscode/src/features/preview.ts
+++ b/editors/vscode/src/features/preview.ts
@@ -58,6 +58,30 @@ type PreviewFileCommandOpts = {
   isNotPrimary?: boolean;
 };
 
+async function launchPreviewWithCompat(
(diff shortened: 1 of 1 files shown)


Myriad-Dreamin:
I'll merge it soon with some minor refactor. in fact I reviewed this feature two weeks ago and I was not satisfied with the performance and limitation: it scans entire workspace and limits number of listed files to 500. I have searched whether we can do better but I didn't find a solution.

PR #2364, opened 2026-01-20
fix: correct `ModuleInclude` syntax identification in collecting links

BlueQuantumx:
Close #2358.

PR #2368, opened 2026-01-21
fix: gain correct file id by joining file id and path when handling LinkTarget::Path

### LINKED ISSUE #2358: Renaming a included file did not change the corresponding typst source
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Platform

Apple Silicon macOS (darwin-arm64, Most Common)

### Editor

VS Cod(e,ium)

### Editor Version

Version: 1.108.1
Commit: 585eba7c0c34fd6b30faac7c62a42050bfbc0086
Date: 2026-01-14T14:55:44.241Z
Electron: 39.2.7
ElectronBuildId: 12953945
Chromium: 142.0.7444.235
Node.js: 22.21.1
V8: 14.2.231.21-electron.0
OS: Darwin arm64 25.2.0

### Describe the bug

Here is a minimal reproduction of the bug:

```typst
// content.typ
---
// base.typ
#include /* position after */ "content.typ"
```

1. try renaming the "content.typ" in base.typ source (using the textDocument/rename LSP feature).

Current behavior:

- In filesystem, the `content.typ` is correctly renamed to `new_name.typ`.
- In `base.typ`, the `"content.typ"` remains unchanged.

Expected behavior:

- In filesystem, current behavior is already good.
- In `base.typ`, the `"content.typ"` is renamed to `new_name.typ`.

### Server Logs

```log
[2026-01-19T08:40:04Z INFO  tinymist::lsp] did open file:///Users/user/Documents/Typst%20Projects/Playground/content.typ
[2026-01-19T08:40:11Z INFO  tinymist::lsp::query] will rename files RenameFilesParams { files: [FileRename { old_uri: "file:///Users/user/Documents/Typst%20Projects/Playground/content.typ", new_uri: "file:///Users/user/Documents/Typst%20Projects/Playground/new_name.typ" }] }
[2026-01-19T08:40:11Z INFO  tinymist_query::will_rename_files] did rename diff: "../new_name.typ"
[2026-01-19T08:40:11Z INFO  tinymist_query::will_rename_files] did rename def_fid: @ws/p0:0.0.0/content.typ
[2026-01-19T08:40:11Z INFO  tinymist_query::will_rename_files] did rename edits: {Url { scheme: "file", cannot_be_a_base: false, username: "", password: None, host: None, port: None, path: "/Users/user/Documents/Typst%20Projects/Playground/base.typ", query: None, fragment: None }: []}
[2026-01-19T08:40:11Z INFO  tinymist::lsp] did open file:///Users/user/Documents/Typst%20Projects/Playground/new_name.typ
[2026-01-19T08:40:11Z INFO  tinymist_project::compiler] /Users/user/Documents/Typst Projects/Playground/base.typ: compilation succeeded in 54.625µs
```

### Browser Logs

Unrelated to browser/preview.

### Additional context

_No response_

BlueQuantumx:
This PR is next to #2364. #2364 is not enough to fix #2358.

In detail, this PR:

- Refactor the `collect_links` function to eliminate unnecessary Option handling.
- Fix the checking on whether a link target is the renamed file, by correcting the joining of target path str with the source file to which a link belongs.
- Correct the fixture indicator in `include_path.typ` and add more include paths to improve fixture coverage.

### DIFF
--- a/crates/tinymist-query/src/rename.rs
+++ b/crates/tinymist-query/src/rename.rs
@@ -162,6 +162,12 @@ pub(crate) fn do_rename_file(
     worker.work(edits)
 }
 
+fn link_path_matches_def(def_fid: TypstFileId, file_id: TypstFileId, path: &str) -> bool {
+    // Compare package and vpath so we avoid allocating a joined file id while
+    // still distinguishing package files that share the same internal path.
+    file_id.package() == def_fid.package() && file_id.vpath().join(path) == *def_fid.vpath()
+}
+
 struct RenameFileWorker<'a> {
     ctx: &'a mut LocalContext,
     def_fid: TypstFileId,
@@ -230,11 +236,9 @@ impl RenameFileWorker<'_> {
         let link_info = get_link_exprs(&ref_src);
         let root = LinkedNode::new(ref_src.root());
         let edits = edits.entry(uri).or_default();
-        // compare vpaths rather than file ids because we don't want to allocate file ids eagerly
-        let def_file_vpath = self.def_fid.vpath();
         for obj in &link_info.objects {
             if !matches!(&obj.target,
-                LinkTarget::Path(file_id, path) if file_id.vpath().join(path.as_ref()) == *def_file_vpath
+                LinkTarget::Path(file_id, path) if link_path_matches_def(self.def_fid, *file_id, path.as_ref())
             ) {
                 continue;
             }
@@ -364,8 +368,12 @@ pub(crate) fn create_change_annotation(
 
 #[cfg(test)]
 mod tests {
+    use std::{path::Path, str::FromStr};
+
     use super::*;
     use crate::tests::*;
+    use tinymist_world::package::PackageSpec;
+    use typst::syntax::VirtualPath;
 
     #[test]
     fn test() {
@@ -393,5 +401,34 @@ mod tests {
 
             assert_snapshot!(JsonRepr::new_redacted(result, &REDACT_LOC));
         });
+    }
+
+    #[test]
+    fn link_path_match_requires_same_package_spec() {
+        let package_v010 = PackageSpec::from_str("@preview/example:0.1.0").unwrap();
+        let package_v011 = PackageSpec::from_str("@preview/example:0.1.1").unwrap();
+        let def_fid = TypstFileId::new(
+            Some(package_v010.clone()),
+            VirtualPath::new(Path::new("/assets/logo.typ")),
+        );
+        let same_package_ref = TypstFileId::new(
+            Some(package_v010),
+            VirtualPath::new(Path::new("/docs/main.typ")),
+        );
+        let other_package_ref = TypstFileId::new(
+            Some(package_v011),
+            VirtualPath::new(Path::new("/docs/main.typ")),
+        );
+
+        assert!(link_path_matches_def(
+            def_fid,
+            same_package_ref,
+            "../assets/logo.typ"
+        ));
+        assert!(!link_path_matches_def(
+            def_fid,
+            other_package_ref,
+            "../assets/logo.typ"
+        ));
     }
 }


crates/tinymist-query/src/analysis/link_expr.rs
```diff
@@ -214,7 +212,7 @@ impl LinkStrWorker {
             return Some(());
         }
 
-        let id = node.span().id()?;
+        let id = node.span().id()?.join(&content);
```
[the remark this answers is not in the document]
BlueQuantumx: /gemini review I have reverted the changes and handled in another way, please review again.

crates/tinymist-query/src/rename.rs
```diff
@@ -232,7 +232,7 @@ impl RenameFileWorker<'_> {
         let edits = edits.entry(uri).or_default();
         for obj in &link_info.objects {
             if !matches!(&obj.target,
-                LinkTarget::Path(file_id, _) if *file_id == self.def_fid
+                LinkTarget::Path(file_id, path) if file_id.join(path) == self.def_fid
```
Myriad-Dreamin: This expression, `file_id.join(path)`, allocates new file ids in global storage. Since typst has low limit of allocated file ids (65536), We tend to avoid allocating file ids whenever possible.

BlueQuantumx: I think typst reuses the file id for same file:
```rust
// in typst-syntax/src/file.rs
impl FileId {
    /// Create a new interned file specification.
    ///
    /// The path must start with a `/` or this function will panic.
    /// Note that the path is normalized before interning.
    #[track_caller]
    pub fn new(package: Option<PackageSpec>, path: VirtualPath) -> Self {
        // Try to find an existing entry that we can reuse.
        //
        // We could check with just a read lock, but if the pair is not yet
        // present, we would then need to recheck after acquiring a write lock,
        // which is probably not worth it.
        let pair = (package, path);
        let mut interner = INTERNER.write().unwrap();
        if let Some(&id) = interner.to_id.get(&pair) {
            return id;
        }
        // ...
    }
}
```

Myriad-Dreamin: Assuming that the language server is working on a workspace which has many documents. when the language server extracts link information from all of the documents:
- previously, with only `LinkTarget::Path(file_id, path)`, the language server doesn't allocate extra file ids unless the language server compiles a related document.
- if we use `file_id.join(path)` here, all of the files in the workspace are allocated with ids eagerly, which is what I'm not in flavor.

BlueQuantumx: To what i knew, renaming is actually a kind of "heavy" work, it intrinsically needs to scan all the links(refs) in all the source files and check whether the link targets to the renamed file. If we don't use file id to identify them, we also need to find a way, e.g., directly composing and comparing the vpaths. The difference in cost may not vary evident. The main concern may be that things about file ids need to gain access locks with the global id manager.

And actually in current implementations, tinymist easily assigns a file id to the vast majority of files. For example, when collecting completion files:
```rust
// tinymist-query/src/analysis/global.rs

/// Get all the source files in the workspace.
pub(crate) fn completion_files(&self, pref: &PathKind) -> impl Iterator<Item = &TypstFileId> {
    let regexes = pref.ext_matcher();
    self.caches
        .completion_files
        .get_or_init(|| {
            if let Some(root) = self.world().entry_state().workspace_root() {
                scan_workspace_files(&root, PathKind::Special.ext_matcher(), |path| {
                    WorkspaceResolver::workspace_file(Some(&root), VirtualPath::new(path))
                })
            } else {
                vec![]
            }
        })
        .iter()
        .filter(move |fid| {
            fid.vpath()
                .as_rooted_path()
                .extension()
                .and_then(|path| path.to_str())
                .is_some_and(|path| regexes.is_match(path))
        })
}
```

The
```rust
scan_workspace_files(&root, PathKind::Special.ext_matcher(), |path| {
    WorkspaceResolver::workspace_file(Some(&root), VirtualPath::new(path))
})
```
allocates almost every file a file id.

However, this behavior may be improved, i'm just using it to convince that we are capable of the cost.

I have no preference between comparing file id or comparing vpaths. Using file id is more concise while using vpaths is faster.

BlueQuantumx: A vpath version may look like this:
```rust
        let def_file_vpath = self.def_fid.vpath();
        for obj in &link_info.objects {
            if !matches!(&obj.target,
                LinkTarget::Path(file_id, path) if file_id.vpath().join(path.as_ref()) == *def_file_vpath
            ) {
                continue;
            }
            if let Some(edit) = self.rename_resource_path(obj, &root, &ref_src) {
                edits.push(edit);
            }
        }
```

PR #2381, opened 2026-01-27
feat: inlay hint for package import version status

QuadnucYard:
Displays inline decorations showing whether a package version is up-to-date, upgradable, or invalid.

image


QuadnucYard:
The snapshots contain l10n messages that require #2380 to correctly display.

PR #2384, opened 2026-01-29
feat(tyck): path typing for include/import sources

hongjr03:
Constrains / sources to  and fixes related typing flow around let initializers and for-loop patterns. Updates post_type_check fixtures/snapshots for include/import paths.

### DIFF
--- a/crates/tinymist-query/src/syntax/expr.rs
+++ b/crates/tinymist-query/src/syntax/expr.rs
@@ -572,7 +572,7 @@ impl ExprWorker<'_> {
         let is_wildcard_import = matches!(typed.imports(), Some(ast::Imports::Wildcard));
 
         let source = typed.source();
-        let mod_expr = self.check_import(typed.source(), true, is_wildcard_import);
+        let mod_expr = self.check_import(source, true, is_wildcard_import);
         crate::log_debug_ct!("checking import: {source:?} => {mod_expr:?}");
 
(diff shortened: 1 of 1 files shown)


crates/tinymist-query/src/analysis/tyck/syntax.rs
```diff
@@ -580,8 +593,61 @@ impl TypeChecker<'_> {
...
-        let _pattern = self.check_pattern_exp(&for_loop.pattern);
+        let iter = self.check(&for_loop.iter);
+        let pattern = self.check_pattern_exp(&for_loop.pattern);
+
+        // todo: This doesn't fully utilize the existing checkers. We have a better way
+        // of implementing this check, add a constraint `array(iter) <: pattern`.
+        // Note: this is not implemented yet in `TypeChecker::constrain`, so we need to
+        // implement similar logic as following checking specific to loop
+        // variables.
+        if matches!(for_loop.pattern.as_ref(), Pattern::Simple(..)) {
```
Myriad-Dreamin, line 604: todo: This doesn't fully utilize the existing checkers. We have a better way of implementing this check, add a constraint `array(iter) <: pattern`. Note: this is not implemented yet in `TypeChecker::constrain`, so we need to implement similar logic in `TypeChecker::constrain` as following check specific to loop variables.

Myriad-Dreamin (approved):

Issue #2126, opened 2025-09-20
Preview theme extension


Myriad-Dreamin:
### Motivation

The preview html can be obtained by installing another extension, and we can specify which extension is used to get a preview html.

### Description

We can specify which extension is used to get a preview html in configuration:

```json
{
  "tinymist.preview.theme.provider": "abc.typst-preview-extension" 
}
```

We can also specify a `html-path` to load a preview HTML at the path:

```json
{
  "tinymist.preview.theme.html-path": "file-path" 
}
```

## Compability checking

A theme provider should check whether it is compatible with the tinymist extension. By default, the tinymist version must be exactly match to a specific one.

## Security consideration

If we are running in an untrusted workspace, the configuration is ignored.

Aitanuqui:
Is this issue a feature request for an html preview?

PR #2426, opened 2026-03-16
Server-side SVG rendering for native editor integrations

dsturnbull:
Adds `--server-svg` and `--strip-svg-glyph-defs` flags to the preview server for native editor integrations that can rasterise SVG directly, without the reflexo vector IR / WASM frontend.

## Motivation

Native editors that already have an SVG rasteriser (e.g. via resvg) can display typst output without pulling in ~250 reflexo rendering crates or shipping a WASM frontend. The tinymist LSP is already running, already watching the file, and already doing incremental compilation with comemo. The only missing piece is a way to get rendered output as SVG instead of vector IR.

This is useful for any editor integration where:
- The editor can rasterise SVG natively or already has resvg/usvg in its dependency tree
- Pulling in the reflexo vector format and its rendering pipeline is too heavy
- A simple text-based WebSocket protocol is preferable to binary vector IR

### `--server-svg`

When passed to `tinymist preview` or via `tinymist.doStartPreview` as a CLI-style argument, the `RenderActor` calls `typst_svg::svg(page)` for each page and sends the result as `WsMessage::Text` instead of the usual `WsMessage::Binary` vector IR.

Multi-page documents send one WebSocket message per page with a header:

```
page:0:5\n<svg class="typst-doc" ...>...</svg>
page:1:5\n<svg class="typst-doc" ...>...</svg>
...
```

The existing incremental/delta rendering path is completely unchanged when the flag is off.

### `--strip-svg-glyph-defs`

Typst SVGs embed every glyph outline as a `<symbol>` inside `<defs id="glyph">`. For a typical document this is 80–90% of the SVG (~1.5MB of ~2MB). The glyph set rarely changes between keystrokes.

When `--strip-svg-glyph-defs` is passed alongside `--server-svg`, the `RenderActor` hashes each page's `<defs id="glyph">` section and strips it from subsequent frames where the hash matches. Per-page hashing so multi-page documents don't cross-contaminate. Reduces per-frame transfer from ~2MB to ~200KB.

Off by default because it changes the SVG content in a way that requires client-side cooperation — the receiving client must cache the defs from the first frame and re-inject them into subsequent frames.

### Files changed

**`crates/typst-preview/src/actor/render.rs`** (+97/-3)
- `render_svg()`: iterates `doc.pages`, renders each with `typst_svg::svg(page)`, prepends `page:{idx}:{total}\n` header
- `strip_cached_glyph_defs()`: per-page hash-based stripping, gated on `self.strip_svg_glyph_defs`
- `cached_glyph_defs_hashes: HashMap<usize, u64>` on `RenderActor`
- Run loop dispatches to `render_svg()` when `self.server_svg` is set, sends each page as a separate channel message

**`crates/typst-preview/src/actor/webview.rs`** (+11/-2)
- Sends `WsMessage::Text` when `server_svg` enabled, `WsMessage::Binary` otherwise

**`crates/typst-preview/src/lib.rs`** (+15)
- `server_svg` and `strip_svg_glyph_defs` fields on `PreviewConfig` and `DataPlane`
- Plumbed through to `RenderActor::new()`

**`crates/tinymist/src/tool/preview.rs`** (+19)
- `--server-svg` and `--strip-svg-glyph-defs` clap arguments on `PreviewArgs`

**`crates/tinymist/src/config.rs`** (+2)
- Defaults for both fields (`false`)

## Example usage

An editor integration would start the preview via the LSP:

```json
{"command": "tinymist.doStartPreview", "arguments": [["--server-svg", "--strip-svg-glyph-defs", "--data-plane-host=127.0.0.1:0", "--task-id=my-preview", "/path/to/file.typ"]]}
```

Then connect to the returned WebSocket port, send `"current"` to trigger the initial render, and receive SVG text messages as the document changes. With `--strip-svg-glyph-defs`, the client caches `<defs id="glyph">...</defs>` from the first frame and injects it into subsequent frames that have it stripped.

The `--task-id` flag allows concurrent previews for different files. `tinymist.doKillPreview` with the same task ID cleans up.

## Performance

On a complex A4 2-column document (536 glyphs, 7704 glyph uses, 5 pages, ~2MB SVG per page): comemo incremental recompile takes ~26ms, leaving ~120ms of frame budget for the client to rasterise and display at typing speed.

## Design notes

- No changes to the default preview behaviour. Both flags are opt-in.
- The vector IR path, browser-based frontend, and all existing preview functionality are untouched.
- The `page:{idx}:{total}\n` header is a simple text prefix, not JSON, to avoid escaping a multi-megabyte SVG string.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6619,6 +6619,7 @@ dependencies = [
  "typst",
  "typst-assets",
  "typst-macros",
+ "typst-svg",
  "typst-timing 0.14.2",
 ]
 
--- a/crates/tinymist/src/config.rs
+++ b/crates/tinymist/src/config.rs
@@ -575,6 +575,8 @@ impl Config {
             refresh_style: self.preview.refresh.clone().unwrap_or(TaskWhen::OnType),
             invert_colors: serde_json::to_string(&self.preview.invert_colors)
                 .unwrap_or_else(|_| "never".to_string()),
+            server_svg: false,
+            strip_svg_glyph_defs: false,
         }
     }
 
--- a/crates/tinymist/src/tool/preview.rs
+++ b/crates/tinymist/src/tool/preview.rs
@@ -113,6 +113,23 @@ pub struct PreviewArgs {
     /// This is hidden from the CLI.
     #[clap(long, hide(true))]
     pub refresh_style: Option<RefreshStyle>,
+
+    /// Render complete SVG server-side instead of streaming vector IR.
+    ///
+    /// When enabled, the WebSocket data plane sends SVG text messages
+    /// instead of binary reflexo vector IR. Intended for native editor
+    /// integrations that can rasterize SVG directly.
+    #[clap(long = "server-svg")]
+    pub server_svg: bool,
+
+    /// Strip unchanged glyph defs from SVGs after the first frame.
+    ///
+    /// Requires --server-svg.  Reduces per-frame WebSocket transfer from
+    /// ~2MB to ~200KB by stripping the `<defs id="glyph">` section when
+    /// it hasn't changed since the previous frame.  The receiving client
+    /// must cache and re-inject the stripped defs.
+    #[clap(long = "strip-svg-glyph-defs")]
+    pub strip_svg_glyph_defs: bool,
 }
 
 impl PreviewArgs {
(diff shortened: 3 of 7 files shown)


Myriad-Dreamin:
Previewing with pure svg has poor performance, so I don't think we should develop for this direction. To improve performance, we can:
- experiment a faster preview method and publish it as another extension (see #2126). If this extension is mature enough, we can switch tinymist's default previewer to the experiment one.
- continue improve #2396, which will be published as an experimental preview extension.

PR #2475, opened 2026-04-05
feat: allow more granular log control

jyn514:
- Change the default for `preview` to `WARN`. Previously it was `INFO`,
  which made it hard to see errors in the terminal.
- Add a new `preview --verbose` flag to allow bringing back the old
  behavior.
- Make verbosity calculation easier to understand by doing it all in one place.
- Allow overriding log levels with `TINYMIST_LOG`. This allows overriding individual modules without having to recompile from source.
- Add docs about logging.

### DIFF
--- a/crates/tinymist/src/log.rs
+++ b/crates/tinymist/src/log.rs
@@ -6,28 +6,17 @@ use super::*;
 
 /// Options for initializing the logger.
 pub struct InitLogOpts {
-    /// Whether the command is transient (e.g., compile).   
-    pub is_transient_cmd: bool,
-    /// Whether the command is a test without verbose output.
-    pub is_test_no_verbose: bool,
+    /// Whether the command should use verbose logging.
+    pub verbose: bool,
     /// Redirects output via LSP/DAP notification.
     pub output: Option<LspClient>,
 }
 
 /// Initializes the logger for the Tinymist library.
-pub fn init_log(
-    InitLogOpts {
-        is_transient_cmd,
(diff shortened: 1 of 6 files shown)


crates/tinymist/src/log.rs
```diff
@@ -59,12 +48,13 @@ pub fn init_log(
...
         .filter_module("tinymist", base_level)
-        .filter_module("tinymist_preview", preview_level)
+        .filter_module("tinymist_preview", base_level)
         .filter_module("typlite", base_level)
         .filter_module("reflexo", base_level)
         .filter_module("sync_ls", base_level)
         .filter_module("reflexo_typst2vec::pass::span2vec", Error)
-        .filter_module("reflexo_typst::diag::console", diag_level)
+        .filter_module("reflexo_typst::diag::console", base_level)
+        .parse_env("TINYMIST_LOG")
```
Myriad-Dreamin: Can we get this argument from clap like this?
```rs
// Get it using clap
    /// Configure log filter of tinymist
    #[clap(long = "log-filter", env = "TINYMIST_LOG")]
    pub log_filter: String,


...
// configure filter using `parse_filters`
        .filter_module("reflexo_typst::diag::console", base_level)
        .parse_filters(args.log_filter)
```

jyn514: good idea, done :)

Myriad-Dreamin (approved):
Thank you! Look good to me.

PR #2494, opened 2026-04-29
fix: switch to gen-lsp-types

### LINKED ISSUE #1542: Consider switching from `lsp-types` to `lspt`
### Motivation

Some nice features will come with lsp 3.18 but `lsp-types` are getting unmaintained with these new features.

### Description

`lsp-types` is not well maintained, and we can consider `lspt`.

### More Examples/Questions

_No response_

ribru17:
[gen-lsp-types](https://github.com/ribru17/gen-lsp-types) is an alternative to the lsp-types crate, with types generated via codegen from the official LSP Metamodel for correctness and completeness.

lsp-types issues fixed in gen-lsp-types:

- https://github.com/gluon-lang/lsp-types/issues/310
- https://github.com/gluon-lang/lsp-types/issues/308
- https://github.com/gluon-lang/lsp-types/issues/284
- https://github.com/gluon-lang/lsp-types/issues/278
- https://github.com/gluon-lang/lsp-types/issues/277
- https://github.com/gluon-lang/lsp-types/issues/260
- https://github.com/gluon-lang/lsp-types/issues/245
- https://github.com/gluon-lang/lsp-types/issues/93

Closes #1542 (as an alternative solution)

### DIFF
--- a/crates/sync-lsp/src/server/dap_srv.rs
+++ b/crates/sync-lsp/src/server/dap_srv.rs
@@ -42,7 +42,8 @@ where
         mut self,
         handler: RawHandler<Args::S, JsonValue>,
     ) -> Self {
-        self.req_handlers.insert(R::COMMAND, Box::new(handler));
+        self.req_handlers
+            .insert(R::COMMAND.into(), Box::new(handler));
         self
     }
 
@@ -54,7 +55,7 @@ where
         handler: fn(&mut Args::S, R::Arguments) -> ScheduleResult,
     ) -> Self {
         self.req_handlers.insert(
-            R::COMMAND,
+            R::COMMAND.into(),
             Box::new(move |s, req| handler(s, from_json(req)?)),
         );
         self
@@ -66,7 +67,7 @@ where
         handler: AsyncHandler<Args::S, R::Arguments, R::Response>,
     ) -> Self {
         self.req_handlers.insert(
-            R::COMMAND,
+            R::COMMAND.into(),
             Box::new(move |s, req| erased_response(handler(s, from_json(req)?))),
         );
         self
@@ -222,7 +223,7 @@ where
 
                 let is_disconnect = method == dapts::request::Disconnect::COMMAND;
 
-                let Some(handler) = self.requests.get(method) else {
+                let Some(handler) = self.requests.get(&method.into()) else {
                     log::warn!("unhandled dap request: {method}");
                     break 'serve_req just_result(Err(method_not_found()));
                 };
@@ -251,7 +252,7 @@ where
                           event,
                           body,
                       }: dap::Event| {
-            let Some(handler) = self.notifications.get(event.as_str()) else {
+            let Some(handler) = self.notifications.get(&event.as_str().into()) else {
                 log::warn!("unhandled event: {event}");
                 return Ok(());
             };
--- a/crates/sync-lsp/src/server/lsp_srv.rs
+++ b/crates/sync-lsp/src/server/lsp_srv.rs
@@ -1,4 +1,4 @@
-use lsp_types::{notification::Notification as Notif, request::Request as Req, *};
+use lsp_types::{Notification as Notif, Request as Req, *};
 
 use super::*;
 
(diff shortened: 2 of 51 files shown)


ribru17:
See https://github.com/rust-lang/rust-analyzer/pull/22115 and https://github.com/wgsl-analyzer/wgsl-analyzer/pull/1090 for other migration considerations

crates/sync-lsp/src/server/lsp_srv.rs
```diff
@@ -351,29 +351,32 @@ where
...
+            handler(s, arguments.unwrap_or_default())
         }
     }
 
     /// Handles an incoming notification.
     pub fn on_notification(&mut self, method: &str, params: JsonValue) -> LspResult<()> {
         let handle = |s, method: &str, params: JsonValue| {
-            let Some(handler) = self.notifications.get(method) else {
+            let Some(handler) = self.notifications.get(&method.to_owned().into()) else {
                 log::warn!("unhandled notification: {method}");
```
Myriad-Dreamin, line 368: Nit: every log of request/notification method introduces an extra allocation. while it doesn't matter much, but I feel it is kind of imperfect:
```rs
impl fmt::Display for LspNotificationMethod {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        let s: String = self.clone().into(); // a clone
        write!(f, "{s}")
    }
}
```

crates/sync-lsp/src/server/dap_srv.rs
```diff
@@ -222,7 +223,7 @@ where
 
                 let is_disconnect = method == dapts::request::Disconnect::COMMAND;
 
-                let Some(handler) = self.requests.get(method) else {
+                let Some(handler) = self.requests.get(&method.to_owned().into()) else {
```
Myriad-Dreamin: I'm not pretty in flavor of converting string enums into `LspRequestMethod` or `LspNotificationMethod` everywhere. And this case is even worse, for a DAP server, every "DapCommandMethod" and "DapEventMethod" are converted to `LspRequestMethod::Custom` and `LspNotificationMethod::Custom` and requires extra allocations. 
Context: LSP and DAP both share a same [base protocol](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/).

What's your idea?

Myriad-Dreamin (commented):
Hi, thanks for the awesome `gen-lsp-types` crate. I really appreciate you providing a replacement for the unmaintained `lsp-types`. I've left two comments, but they shouldn't block merging.

crates/sync-lsp/src/server/lsp_srv.rs
ribru17: Good point. I think the best thing to do is to move from a custom variant of `Cow<'static, str>` to just `&str`, where the reference lifetime is a parameterized `'a`. That way, the enum becomes `Copy` and can be constructed from any `&str` without any allocations, while still allowing exhaustive enum-style pattern matching. What do you think?

crates/sync-lsp/src/server/dap_srv.rs
ribru17: Good point... maybe it's better to remove the typing guarantee and keep these keys as &str. This pain point will be largely mitigated by implementing the suggestion from my previous comment, but the `Custom` dap request/notification variants will still be present since the LSP metamodel is unaware of the DAP specification.

ribru17:
I'm drafting another release of the library which contains a metamodel update (supporting new markdown diagnostic messages) and allows for request/notification enums to be created without ownership, meaning no unnecessary allocations. Will mark this PR as ready once the release is cut and everything seems to be working :+1:

ribru17:
I've pushed a new version of the library, which should remove the need for the allocations as mentioned above. Also, the request/notification method newtype is now `Copy` as well! PTAL and let me know your thoughts :)

ribru17:
@Myriad-Dreamin I've resolved the conflicts for this change and bumped to the latest library version (in case you're still interested in this change)

Myriad-Dreamin:
@ribru17 I didn't forget this PR. we are just processing bump the main upstream typst from v0.14 to v0.15. This just happens once a year, :).

ribru17:
No worries or rush- that's exciting!

Issue #1929, opened 2025-07-26
Keep vertical position stable when resizing the preview window


hooyuser:
### Motivation

Currently, when I'm viewing a document in the preview panel and I drag the left side of the panel to resize it, the content often scrolls up or down unexpectedly. This forces me to find my place in the document again and again, especially when I'm trying to make small adjustments of the preview panel.

### Description

It would be awesome if the preview could lock the vertical position while the panel is being resized horizontally. Here are two images to illustrate this behavior.

- before dragging the left side of the panel
     <img width="423" height="895" alt="Image" src="https://github.com/user-attachments/assets/55ce433a-5d4b-4dc2-837e-7dcbaa1d0545" />

- after dragging the left side of the panel
    <img width="890" height="902" alt="Image" src="https://github.com/user-attachments/assets/639052bb-c6a5-4adf-a3d3-194f0f6ca736" />

To describe it more precisely: imagine the document has absolute coordinates, with x and y representing the horizontal and vertical axis respectively. Before I start dragging to resize the window, the very top of the visible viewport corresponds to a specific coordinate, `y1`, in the document. My ideal behavior is that throughout the entire drag/resize operation, the top of the viewport remains aligned with `y1`. The content could still resize or reflow, but we wouldn't lose the vertical spot.

This would make resizing feel much more stable and predictable.

### More Examples/Questions

Building on this core idea, it would be even more powerful to support two distinct resizing modes, similar to what some advanced PDF viewers offer:
 
1. "Fit to Width" Mode (with vertical lock)

   Behavior: When resizing the panel, the top of the viewport stays locked to its absolute y-coordinate in the    document, as described above.

   Scaling: The document content scales (zooms in or out) so that its width always fits the new panel width. This is essentially an improved version of the current behavior, just with the vertical jumping fixed.

2. "Absolute Size" Mode (with vertical lock)

   Behavior: The top of the viewport is also locked vertically.

   Scaling: The document does not scale. The zoom level remains at its current setting (e.g., 100%).

   Horizontal Move: The horizontal position would adjust intelligently based on what's visible, according to the following logic:
 ```
IF the right side of the document is clipped by the viewbox THEN
    // before moving
    let current_x_left = current document's absolute x-coordinate at the left edge of the viewbox
    // start moving
    when moving, keep aligning the left edge of the viewbox with `current_x_left`
    // Let the clipped content on the right enter the viewbox
ELSE
[...]

PR #2495, opened 2026-05-02
fix: restore the scroll position when resizing the preview panel

hooyuser:
This PR fixes the bug reported in #1929 , preserving a stable scroll anchor during resize bursts. It does not implement the mode switch between`Fit to Width` and `Absolute Size` proposed in #1929 .

- Reuses the initial scroll anchor for 600ms during continuous resizing, avoiding frame-by-frame anchor recalculation and cumulative jitter.
- Restores scroll position by tracking the anchor’s relative position within the current page (`pageLocalY`), ensuring the scroll position immune to `heightMargin`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6339,7 +6339,7 @@ dependencies = [
  "strum",
  "sync-ls",
  "temp-env",
- "tinymist-assets 0.14.16-rc1",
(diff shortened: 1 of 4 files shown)


Myriad-Dreamin (approved):
LGTM

PR #2569, opened 2026-06-14
fix(typlite): adapt paragraph and line breaks for typst 0.15

Myriad-Dreamin:
## Summary
- Adjust Typlite paragraph conversion to flush the inline buffer on `<p>` elements so paragraph boundaries survive Typst 0.15 HTML output.
- Disable the old linebreak/parbreak conversion path that conflicts with the current Typst-generated HTML shape.
- Keep the corresponding snapshot updates in the snapshot PR for unified review.

### DIFF
--- a/crates/typlite/src/markdown.typ
+++ b/crates/typlite/src/markdown.typ
@@ -228,7 +228,7 @@
     },
   )
 
-  show linebreak: it => if-not-paged(it, md-linebreak)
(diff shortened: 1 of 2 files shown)


Myriad-Dreamin:
Closing this PR because its Typlite commit is already included in stable/v0.15.0 via #2570. Keeping it open after #2570 would make the stacked branch state misleading.

PR #2671, opened 2026-07-22
fix(query): don't invert completion range on spread with extra dots

### LINKED ISSUE #2669: tinymist panicked when calling Ord::clamp
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Platform

x64 Windows (win32-x64, Most Common)

### Editor

Neovim

### Editor Version

NVIM v0.11.7
Build type: Release
LuaJIT 2.1.1741730670

### Describe the bug

tinymist panicked with the following assertion failure:
```
thread 'tokio-rt-worker' (752) panicked at /rustc/ded5c06cf21d2b93bffd5d884aa6e96934ee4234\\library\\core\\src\\cmp.rs:1090:9:
assertion failed: min <= max
```

when looking up where that assertion failed, even though it's not exact, I found the exact same [call to `assert!`](https://doc.rust-lang.org/src/core/cmp.rs.html#1103) in Ord::clamp.

### Server Logs

[tinymist.log](https://github.com/user-attachments/files/30253624/tinymist.log)

### Browser Logs

[tinymist_browser.log](https://github.com/user-attachments/files/30253684/tinymist_browser.log)

### Additional context

This seems to happen whenever I spread an expression with three or more dots (I forget sometimes) instead of the correct two.

SAY-5:
Typing a spread with an extra dot, e.g. `#f(...args)`, panics the server with `assertion failed: min <= max` from `Ord::clamp` in `insert_range_of`. The third dot is classified as a dot access whose target ends up being the whole call, so `field.offset()` lands past the cursor and the replace range `from..cursor` is inverted. Clamp `from` to the cursor, and add a completion fixture that panics without this change.

Fixes #2669

### DIFF
--- a/crates/tinymist-query/src/analysis/completion.rs
+++ b/crates/tinymist-query/src/analysis/completion.rs
@@ -702,7 +702,10 @@ impl CompletionPair<'_, '_, '_> {
                 let target = var.accessed_node()?;
                 let field = var.accessing_field()?;
 
-                self.cursor.from = field.offset(&self.cursor.source)?;
+                // The field offset can land past the cursor on incomplete input
+                // like `f(...x)`, where the classified dot access is the whole
+                // call. Keep `from <= cursor` so the replace range stays valid.
+                self.cursor.from = field.offset(&self.cursor.source)?.min(self.cursor.cursor);
 
                 self.doc_access_completions(&target);
                 return Some(());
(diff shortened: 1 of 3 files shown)


SAY-5:
bumping this. the completion range ends up inverted when the spread has an external argument, so the returned edit range has start > end.

PR #2685, opened 2026-08-05
Add a codeLens to pin the preview for clients without runCodeLens

leiflm:
## Summary

The background/browsing preview (`preview.background.enabled` /
`tinymist.doStartBrowsingPreview`) follows whichever document the server most
recently saw opened (`ServerState::focus_main_file`). Opening a second `.typ`
file re-points the preview to it, and closing that file again does **not**
hand the preview back — edits to the original document then stop producing
any preview updates at all, with no way to recover short of restarting the
server.

`tinymist.pinMain` already exists to fix exactly this, but reaching it
requires the client to send an arbitrary `workspace/executeCommand`. VS Code
gets there through editor-title buttons the extension registers itself
(`tinymist.pinMainToCurrent` / `tinymist.unpinMain` in
`editors/vscode/src/extension.ts`). Clients that only implement generic LSP
UI — Zed, Helix, Neovim, etc. — have no such button and no way to invoke the
command, so the preview simply breaks for them the first time a second file
is opened, with no recourse.

This showed up concretely while debugging why tinymist's preview is
unreliable in Zed with a large, multi-file project.

## Change

**Commit 1** adds a **"📌 Pin Preview to This File" / "📌 Unpin Preview"** code
lens whose command is `tinymist.pinMain` directly, with the toggle state
computed from `pinning_by_user` / `focusing` in `ServerState::code_lens`.

Unlike the existing Profile/Preview/Export/More lenses, this one is **not**
gated behind `support_client_codelens`. Those lenses all route through
`tinymist.runCodeLens`, which only does something because the VS Code
extension intercepts that command client-side before it would reach the
server — there is no server-side handler for `runCodeLens` at all, so on a
client that forwards it as a plain `workspace/executeCommand` it is a no-op.
`tinymist.pinMain` has no such indirection: it's a real command the server
already implements, so the lens works for any standards-compliant LSP
client.

**Commit 2** adds dynamic registration support for codeLens (mirroring the
existing pattern for semantic tokens / document formatting), since Zed
declares `textDocument.codeLens.dynamicRegistration: true`.

## ⚠️ Update after testing against real Zed

I initially assumed the lens alone would be enough, and added the dynamic
registration commit hoping it would unlock things further after finding Zed
never queried `textDocument/codeLens` despite the lens being enabled with
`"code_lens": "on"` / `"menu"`.

**It didn't.** Confirmed via Zed's own `dev: open language server logs` RPC
trace (opened *before* the language server handshake, so the whole session
is captured): Zed 1.13.2 never sends `textDocument/codeLens` for tinymist,
with or without dynamic registration. Other requests (completion, hover,
documentHighlight) work fine on the same connection, so this isn't a broken
connection — codeLens specifically is never queried.

So, as it stands, **this does not currently unlock pinning in Zed**. I'm
still submitting both commits because:
- The lens itself (commit 1) is correct, generically useful LSP surface for
  any client that does support codeLens (confirmed via direct protocol
  testing — see below), and costs nothing for clients that don't.
- The dynamic registration commit (commit 2) is a real, independently
  correct protocol conformance improvement, symmetric with how tinymist
  already handles this exact situation for semantic tokens and document
  formatting. It's plausible some other client needs it even if Zed doesn't
  currently benefit.

I'd guess the Zed-side gap is either a bug in Zed's codeLens support for
extension-provided (WASM) language servers specifically, or something more
fundamental about how it decides which servers to query for lenses. I don't
have Zed's source handy to dig further; happy to file an issue there if that
seems like the right next step, or to close this PR if you'd rather not
carry code with no currently-working consumer.

## Testing

- `cargo build --release -p tinymist-query -p tinymist -p tinymist-cli`
- `cargo clippy --release -p tinymist-query -p tinymist --no-deps` — clean
- `cargo fmt -p tinymist-query -p tinymist -- --check` — clean
- `./target/release/tinymist-l10n --kind rs --dir ./crates --output ./locales/tinymist-rt.toml`
  to extract the new locale keys (added `zh` translations by hand alongside
  the generated `en` ones)
- End-to-end, against the built binary, via a minimal Python LSP client
  driving the exact message sequence a real client sends:
  1. `didOpen` a large document with `preview.background.enabled: true`.
  2. `textDocument/codeLens` → confirmed the response includes
     `{"title": "📌 Pin Preview to This File", "command": "tinymist.pinMain", "arguments": [<path>]}`.
  3. `workspace/executeCommand` with that exact command/arguments (what any
     LSP client sends when the lens is clicked) → succeeds.
  4. `textDocument/codeLens` again → now reads "📌 Unpin Preview" with
     `arguments: [null]`.
  5. Opened a **second** `.typ` file, then edited the first again: before
     this change the preview goes permanently silent at this point (0 bytes
     on the data-plane socket); after pinning, the first document keeps
     receiving incremental preview updates as expected.
  6. Repeated with `textDocument.codeLens.dynamicRegistration: true`
     declared: confirmed `codeLensProvider` is correctly omitted from the
     static capabilities and a `client/registerCapability` request for
     `textDocument/codeLens` is sent after `initialized`; `textDocument/codeLens`
     still answers correctly afterward.
  7. Repeated without declaring it: confirmed `codeLensProvider` is still
     statically declared as before (no regression for existing clients).
- Live against real Zed 1.13.2 (see above): confirmed the connection,
  binary, and other LSP features all work normally; confirmed via RPC trace
  that codeLens specifically is never requested, with or without dynamic
  registration.

## Notes for maintainers

I didn't open an issue first per CONTRIBUTING.md's guidance for new
features, since this is small, additive, reuses an existing command
unchanged, and directly fixes broken behavior rather than adding new
capability surface — but given the Zed result above, I'm very open to
discussion on whether this is worth carrying at all right now, or should
wait until there's a client that actually benefits from it.

Assisted-by: Artificial Intelligence

### DIFF
--- a/crates/tinymist-query/src/code_lens.rs
+++ b/crates/tinymist-query/src/code_lens.rs
@@ -10,6 +10,9 @@ use crate::{SemanticRequest, prelude::*};
 pub struct CodeLensRequest {
     /// The path of the document to request for.
     pub path: PathBuf,
+    /// Whether this document is the one currently pinned as the main file by
+    /// the user (via `tinymist.pinMain`).
+    pub is_pinned_here: bool,
 }
 
 impl SemanticRequest for CodeLensRequest {
@@ -21,6 +24,38 @@ impl SemanticRequest for CodeLensRequest {
         let mut res = vec![];
 
         let doc_start = ctx.to_lsp_range(0..0, &source);
+
+        // Unlike the lenses below, `tinymist.pinMain` is a real, directly
+        // executable LSP command, so this lens works for every client and is
+        // not gated behind `support_client_codelens`. It is the only way for a
+        // client that cannot run arbitrary editor commands (e.g. Zed, Helix,
+        // Neovim) to keep the background/browsing preview on one document:
+        // that preview otherwise follows whichever document was most recently
+        // opened, and closing a second document does not hand it back. See
+        // `ServerState::focus_main_file`.
+        let pin_title = if self.is_pinned_here {
+            tinymist_l10n::t!("tinymist-query.code-lens.unpinMain", "📌 Unpin Preview")
+        } else {
+            tinymist_l10n::t!(
+                "tinymist-query.code-lens.pinMain",
+                "📌 Pin Preview to This File"
+            )
+        };
+        let pin_arg = if self.is_pinned_here {
+            JsonValue::Null
+        } else {
+            self.path.display().to_string().into()
+        };
+        res.push(CodeLens {
+            range: doc_start,
+            command: Some(Command {
+                title: pin_title.to_string(),
+                command: "tinymist.pinMain".to_string(),
+                arguments: Some(vec![pin_arg]),
+            }),
+            data: None,
+        });
+
         let mut doc_lens = |title: &str, args: Vec<JsonValue>| {
             if !ctx.analysis.support_client_codelens {
                 return;
--- a/crates/tinymist/src/config.rs
+++ b/crates/tinymist/src/config.rs
@@ -955,6 +955,10 @@ pub struct ConstConfig {
     pub doc_line_folding_only: bool,
     /// Allow dynamic registration of document formatting.
     pub doc_fmt_dynamic_registration: bool,
+    /// Allow dynamic registration of code lenses. Some clients (e.g. Zed)
+    /// declare this but never query `textDocument/codeLens` when the server
+    /// only advertises `codeLensProvider` statically in `initialize`.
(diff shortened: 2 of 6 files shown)


leiflm:
Closing for now: this fix hasn't actually solved the original problem in my own Zed setup yet (see PR description — Zed 1.13.2 never queries `textDocument/codeLens` regardless of static or dynamic registration). Withdrawing until I've confirmed the underlying preview/LSP issue is actually resolved for real day-to-day use. Will reopen or resubmit if that changes.

PR #2709, opened 2026-09-11
fix: validate is_dir after applying export extension to output path

benzeneringlq:
Change `is_dir` to execute after `with_extension`.

# Reproduction
`main.typ`
export pdf fail if there is a directory named  `main`
`crates/tinymist/src/task/export.rs:377:13: ExportTask(ExportPdf(ExportPdfTask { export: ExportTask { when: Never, output: Some(PathPattern("")), transform: [] }, pages: None, pdf_standards: [], no_pdf_tags: false, creation_timestamp: None })): output path is a directory: "/Users/test/projects/test/main"`


Myriad-Dreamin:
Good catch!

